PR #28, opened 2019-09-20
WIP py37 support

cowlicks:
Same warnings as other py37 PR's

todos:
* the ordered map submodule stuff should be removed. It is a dependenciy of a different "Library" library. It should prob be bundled with that library.
* I renamed some include paths of dependencies to be consistent with the directory structure of the upstream projects. This is so we don't have to create our own directory structure for their dependency.
* SOFA isn't handled here. I installed it locally on my system. We should add it as a git submodule and use ADD_SUBDIRECTORY to manage it.
* Cspice doesn't provide a git repo, or way to build it. It isn't included here, but I had it in thirdparty/. Whatever solution we come up with for managing it, can probably be applied to GTEngine.


lucas-bremond:
Open Space Toolkit 0.4.0 now supports Python 3.7:
https://github.com/open-space-collective/open-space-toolkit-physics/releases/tag/0.4.0

PR #57, opened 2021-01-30
Fix TEME/ITRF angular velocity /2

robinpdm:


test/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/TEME.test.cpp
```diff
@@ -97,13 +97,14 @@ TEST (OpenSpaceToolkit_Physics_Coordinate_Frame_Providers_TEME, GetTransformAt)
 
                 // Angular velocity test
 
-                const Vector3d referenceAngularVelocity_TEME_ITRF_in_TEME = { referenceRow[5].accessReal(), referenceRow[6].accessReal(), referenceRow[7].accessReal() } ;
+                const Vector3d referenceAngularVelocity_ITRF_TEME = { referenceRow[5].accessReal(), referenceRow[6].accessReal(), referenceRow[7].accessReal() } ;
```
lucas-bremond, line 100: The angular velocity notation should be preserved, as both the velocity value and projection frame are referenced:

For example:

```
w_B_A_in_A
```

is different from:

```
w_B_A_in_B
```

even though it represents the same angular velocity (same vector magnitude), but projected in two different frames.

lucas-bremond: As a result:

```
w_B_A_in_A = - w_A_B_in_A
```

robinpdm: Good point! Fixed it.

PR #78, opened 2022-07-23
fix: add suggestions and improvements

lucas-bremond:


bindings/python/test/coordinate/frame/providers/iers/test_manager.py
```diff
@@ -76,12 +81,7 @@ def test_get_bulletin_a_array_success (self, manager: Manager):
 
     def test_get_bulletin_a_at_success (self, manager: Manager):
 
-        try:
-            bulletin_a: BulletinA = manager.get_bulletin_a_at(Instant.now() - Duration.days(8.0))
-        except RuntimeError:
-            manager.reset()
-            manager.clear_local_repository()
-            bulletin_a: BulletinA = manager.get_bulletin_a_at(Instant.now() - Duration.days(5.0))
+        bulletin_a: BulletinA = manager.get_bulletin_a_at(Instant.now() - Duration.days(7.0))
```
Derollez, line 84: I think that one might break at a specific time of the run, @apaletta3 do you confirm? (might be okay, but just checking)

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Gravitational/Earth.cpp
```diff
@@ -40,7 +40,7 @@ inline void                     OpenSpaceToolkitPhysicsPy_Environment_Gravitatio
 
         enum_<Earth::Type>(earth_class, "EarthType")
 
-            .value("spherical", Earth::Type::Spherical)
+            .value("Spherical", Earth::Type::Spherical)
```
Derollez, line 43: Agreed, @apaletta3 we had discussed this, nah?

bindings/python/test/coordinate/frame/providers/iers/test_manager.py
lucas-bremond: So in this diff, I've basically reset the changes introduced by @apaletta3 in his PR.

Since the tests are passing on `master`, why the need to change them (and remove the context cleanup in the fixture)?

Derollez: > So in this diff, I've basically reset the changes introduced by @apaletta3 in his PR.

Yup yup and agreed that it passes on master but I believe at a certain point in time it might not (just because of the issue/release date of the bulletin A). And that's not a big deal because there might only be one day or a few hours where that test based on `now` does not go through, just wanted to point that out ^^ (aligned on the context clean-up in the fixture though)

lucas-bremond: What I fail to understand though: would the proposed changes in the base PR actually make the test pass 100% of the time?

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Gravitational/Earth.cpp
apaletta3: yes we had, sounds good

bindings/python/test/coordinate/frame/providers/iers/test_manager.py
```diff
@@ -23,23 +23,28 @@
...
-from ostk.physics.coordinate.frame.providers.iers import Manager
-from ostk.physics.coordinate.frame.providers.iers import BulletinA
-from ostk.physics.coordinate.frame.providers.iers import Finals2000A
+import ostk.physics as physics
+
+################################################################################################################################################################
+
+Manager = physics.coordinate.frame.providers.iers.Manager
+BulletinA = physics.coordinate.frame.providers.iers.BulletinA
+Finals2000A = physics.coordinate.frame.providers.iers.Finals2000A
```
apaletta3, line 32: Why is this necessary? doesn't importing the whole ostk physics module take more time than just importing the necessary files?

apaletta3 (changes requested):

bindings/python/test/environment/gravitational/earth/test_manager.py
```diff
@@ -1,14 +1,13 @@
 ################################################################################################################################################################
 
 # @project        Open Space Toolkit ▸ Physics
-# @file           bindings/python/test/environment/gravitational/earth/test_earth_manager.py
+# @file           bindings/python/test/environment/gravitational/earth/test_manager.py
```
apaletta3, line 4: tbh I think there is a lot of naming problems in ostk cause all Earths have the same name, certain Celestials have the same name etc... don't we want to move in a direction where classes that do things for different other classes have different names? test_iers_manager or test_earth_manager is not much more difficult to write out and it improves clarity. Same thing with having an EarthGravitationalModel instead of just an Earth, because we end up referring to it as an EarthGravitationalModel in discussions on slack anyways.

apaletta3 (approved):
Thanks for the modifications to my PR! 
Question: Is the Pimpl construct that you use in the gravitational earth model Earth.cpp necessary for performance reasons or does it have a significant advantage? Cause in it of itself it does make the code more complicated to read and understand what's going on. Lmk.

lucas-bremond:
So the rationale behind using Pimpl (which is a common C++ pattern) is because we do not want to expose the implementation to the library user.

If for example we didn't use Pimpl and instead had a GeographicLib object pointer used as our Model class (private) member, then we'd have to `include` GeographicLib inside the model **header** file because the pointer type needs a definition, now adding a requirement that the lib user also has to install the GeographicLib headers in order to compile OSTk.

By using Pimpl, all a OSTk library user has to do is to install the OSTk headers and the compiled shared library binaries.

bindings/python/test/environment/gravitational/earth/test_manager.py
lucas-bremond: But the directory nesting + namespaces provides the necessary context, why bloating the file and class names?

bindings/python/test/coordinate/frame/providers/iers/test_manager.py
```diff
@@ -23,23 +23,28 @@
...
 
 ################################################################################################################################################################
 
 @pytest.fixture
 def manager () -> Manager:
 
     manager = Manager.get()
-    manager.reset()
```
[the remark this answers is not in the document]
lucas-bremond: Isn't that what was done after the yield in the fixture?

bindings/python/test/coordinate/frame/providers/iers/test_manager.py
lucas-bremond: Fair, your approach worked too, happy to revert

bindings/python/test/environment/gravitational/earth/test_manager.py
apaletta3: I see where you're coming from, but to me Earth vs EarthGravitationalModel doesn't sound like bloat and makes the whole library easier to navigate and program.

apaletta3 (approved):

apaletta3:
Okay I see, thanks for the explanation behind why Pimpl is used in ostk, makes sense!

I'm pushing a commit with reverts to my original changes for test_manager.py only. The try catch block is necessary to allow the get_bulletin_at() method to work 100% of the time due to a quirk in how often the bulletin A is posted online and its EOP correction parameters timespan. An update was posted on July 21, which covers from July 15th at 00:00 to July 21st at 00:00, not July 21st at 23:59, which is only 6x24 hours of coverage. Therefore, if you try and do get_bulletin_at(now - 7 days) on July 21st it will not work and error, but if you try from July 22nd to July 27th it will. Then again on July 28th it will fail cause there is no data for the day of July 21st (nothing between 00:01 and 23:59) before the bulletin is updated, and once the bulletin is updated sometime during the day of the 28th there will again be no data for July21st because the bulletin will start at July 22 00:00. This isn't a problem with get_bulletin_at since the get_polar_motion_at method has more logic to handle this edge case and interpolate forwards in time for requests into the future or use to the finals2000A log for requests more than 6 days in the past, so get_polar_motion_at will not break. However, the unit test I changed in this PR tests get_bulletin_at specifically, which will break every 7 days so this is why a try catch block around it is necessary.

PR #85, opened 2022-10-18
fix: remove this as dependency of SPICE Engine lambda

lucas-bremond:
- Remove `this` from SPICE Engine lambda function (as this was a bad practice)
- Remove unused `std::cout` from codebase


vishwa2710 (approved):
Looks good. For my knowledge, why is using 'this' in lambdas bad practice?

lucas-bremond:
@vishwa2710 It's not "using `this` in a lambda" which is a bad-practice per-se, it's the fact that this lambda is returned (as part of a transform provider) to the outside word, which could lead to segfaults if the object that created it (pointed by `this`) is destroyed.

Derollez (approved):

Derollez:
@lucas-bremond Just for my understanding, did you catch that while working on `GenerateFrame` used in `Satellite` in https://github.com/open-space-collective/open-space-toolkit-simulation/pull/6?

PR #88, opened 2022-10-31
feat: add J2000 frame provider

lucas-bremond:


src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/J2000.cpp
```diff
@@ -0,0 +1,149 @@
...
+            double gamb, phib, psib, epsa ;
+
+            iauPfw06(DJM0, DJM00, &gamb, &phib, &psib, &epsa) ;
+            iauFw2m(gamb, phib, psib, epsa, rb) ;
+
+            break ;
+
+        }
+
+        default:
```
vishwaLoft, line 119: Since we're not supporting the IAU_2000B case should we remove it/comment it out for now from the above files?

vishwaLoft (approved):
This is great. Just 1 small comment.

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/J2000.cpp
lucas-bremond: I actually have an upcoming PR which will make use of IAU 2000B for TOD calculation :)

vishwaLoft: Ah gotcha. That makes sense!

vishwaLoft (approved):

PR #92, opened 2023-01-20
fix: time interval intersects method binding

Derollez:
- Fix intersects method binding. 
- Add args in python binding. 
- Refactor unit tests using pytest fixtures. 
- Add tools to mount and link with ostk dependency repositories

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Interval.cpp
+++ b/bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Interval.cpp
@@ -26,7 +26,15 @@ inline void                     OpenSpaceToolkitPhysicsPy_Time_Interval     (   
 
     class_<Interval> interval_class(aModule, "Interval") ;
 
-    interval_class.def(init<const Instant&, const Instant&, const Interval::Type&>())
+    interval_class
+
+        .def
+        (
+            init<const Instant&, const Instant&, const Interval::Type&>(),
+            arg("start_instant"),
+            arg("end_instant"),
+            arg("type")
+        )
 
         .def(self == self)
         .def(self != self)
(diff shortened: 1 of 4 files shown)


tools/development/start.sh
```diff
@@ -4,61 +4,108 @@
 
 # @project        Open Space Toolkit ▸ Physics
 # @file           tools/development/start.sh
-# @author         Lucas Brémond <lucas@loftorbital.com>
-# @license        Apache License 2.0
+# @author         Remy Derollez <remy@loftorbital.com>
+# @license        Copyright © 2018-2022 Loft Orbital Solutions Inc.
```
robinpdm, line 8:
```suggestion
# @license        Apache License 2.0
```

bindings/python/test/time/test_interval.py
```diff
@@ -2,137 +2,219 @@
...
-    assert interval.intersects(interval) is not None
+    return Instant.J2000() + Duration.minutes(3.0)
 
 ################################################################################################################################################################
 
-def test_interval_contains_instant ():
+class TestInterval:
 
-    interval = Interval(Instant.J2000(), Instant.J2000() + Duration.minutes(1.0), Interval.Type.Closed)
+    def test_interval_constructor (self):
```
robinpdm, line 85: Detail: did we drop the tested class name in the test names already elsewhere (in this project)?

```cpp
class TestClass:
      def test_constructor (self):
```
instead of:
```cpp
class TestClass:
      def test_class_constructor (self):
```

If we're in an in-between state, let's realign everything later

robinpdm (approved):

bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Interval.cpp
```diff
@@ -36,9 +44,9 @@ inline void                     OpenSpaceToolkitPhysicsPy_Time_Interval     (
 
         .def("is_defined", &Interval::isDefined)
         .def("is_degenerate", &Interval::isDegenerate)
-        .def("intersects", &Interval::intersects)
-        .def("contains_instant", +[] (const Interval& anInterval, const Instant& anInstant) -> bool { return anInterval.contains(anInstant) ; })
-        .def("contains_interval", +[] (const Interval& anInterval, const Interval& anOtherInterval) -> bool { return anInterval.contains(anOtherInterval) ; })
+        .def("intersects", +[] (const Interval& anInterval, const Interval& anotherInterval) -> bool { return anInterval.intersects(anotherInterval) ; }, arg("interval"))
```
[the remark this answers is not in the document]
Derollez: Yes, basically Interval (in physic) inherits from Interval<Instant> from Math. And `intersects` is defined at parent level.Some casting is necessary

bindings/python/test/time/test_interval.py
Derollez: Yeah, we are in between (as you can see not everything as arg, not all the tests use fixtures), we are re-aligning as we go

PR #93, opened 2023-02-11
feat: add timezone for UTC IS08601 toString

vishwa2710:


bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Instant.cpp
```diff
@@ -55,7 +56,8 @@ inline void                     OpenSpaceToolkitPhysicsPy_Time_Instant      (
         .def("get_modified_julian_date", &Instant::getModifiedJulianDate)
         .def("get_leap_second_count", &Instant::getLeapSecondCount)
         .def("to_string", +[] (const Instant& anInstant) -> String { return anInstant.toString() ; })
-        .def("to_string", +[] (const Instant& anInstant, const Scale& aScale) -> String { return anInstant.toString(aScale) ; })
+        .def("to_string", +[] (const Instant& anInstant, const Scale& aScale) -> String { return anInstant.toString(aScale) ; }, arg("scale"))
+        .def("to_string", +[] (const Instant& anInstant, const Scale& aScale, const DateTime::Format& aDateTimeFormat) -> String { return anInstant.toString(aScale) ; }, arg("scale"), arg("date_time_format"))
```
lucas-bremond, line 60: Two comments on this:

1. There's a better way now to overload bindings with Pybind11, see [docs](https://pybind11.readthedocs.io/en/stable/classes.html#overloaded-methods). Search for the OSTk codebase for usage, it's actually simpler than using lambdas (this approach dates back from when we were using Boost Python instead).
2. The second overload doesn't actually use `aDateTimeFormat`, is this expected?

src/OpenSpaceToolkit/Physics/Time/Instant.cpp
```diff
@@ -374,14 +374,20 @@ Int64                           Instant::getLeapSecondCount                 ( )
...
     {
         throw ostk::core::error::runtime::Undefined("Instant") ;
     }
 
+    if ((aTimeScale == Scale::UTC) && (aDateTimeFormat == DateTime::Format::ISO8601))
+    {
+        return this->getDateTime(aTimeScale).toString(aDateTimeFormat) + "Z" ;
+    }
+
     return this->getDateTime(aTimeScale).toString() + " [" + StringFromScale(aTimeScale) + "]" ;
```
lucas-bremond, line 391: What about something like:

```suggestion
    if ((aDateTimeFormat != DateTime::Format::Standard) && (aTimeScale == Scale::UTC))
    {
        return this->getDateTime(aTimeScale).toString(aDateTimeFormat) + "Z" ;
    }
    
    return this->getDateTime(aTimeScale).toString(aDateTimeFormat) + " [" + StringFromScale(aTimeScale) + "]" ;
```

This way the newly added `aDateTimeFormat` arg can be used in all cases.

test/OpenSpaceToolkit/Physics/Time/Instant.test.cpp
```diff
@@ -1216,6 +1216,7 @@ TEST (OpenSpaceToolkit_Physics_Time_Instant, ToString)
         for (auto const& scale : scales)
         {
 
+
```
lucas-bremond, line 1219: TBR?

bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Instant.cpp
robinpdm, line 60:
```suggestion
        .def("to_string", +[] (const Instant& anInstant, const Scale& aScale, const DateTime::Format& aDateTimeFormat) -> String { return anInstant.toString(aScale, aDateTimeFormat) ; }, arg("scale"), arg("date_time_format"))
```

robinpdm: Add Python test?

src/OpenSpaceToolkit/Physics/Time/Instant.cpp
robinpdm: Or
```suggestion
    if ((aDateTimeFormat == DateTime::Format::ISO8601) && (aTimeScale == Scale::UTC))
    {
        return this->getDateTime(aTimeScale).toString(aDateTimeFormat) + "Z" ;
    }
    
    return this->getDateTime(aTimeScale).toString(aDateTimeFormat) + " [" + StringFromScale(aTimeScale) + "]" ;
```
since the `Z` suffix is ISO 8601 specific? No?

bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Instant.cpp
vishwa2710: Excellent. Agreed that's a better syntax, and yes I totally brain farted on not using the variable.

src/OpenSpaceToolkit/Physics/Time/Instant.cpp
vishwa2710: Yeah that's correct. I believe this works well and addresses Lucas' point.

lucas-bremond (approved):

PR #96, opened 2023-05-03
fix: add suggestions

lucas-bremond:


include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Exponential.hpp
```diff
@@ -51,26 +53,35 @@ class Exponential : public Model
...
+
         /// @brief              Get the atmospheric density value at a given position and instant
         ///
         /// @param              [in] aLLA A position, expressed as latitude, longitude, altitude [deg, deg, m]
         /// @param              [in] anInstant An instant
         /// @return             Atmospheric density value [kg.m^-3]
 
-        virtual Real            getDensityAt                                (   const   LLA&                        aLLA,
-                                                                                const   Instant&                    anInstant                                   ) const override ;
+        Real                    getDensityAt                                (   const   LLA&                        aLLA,
```
kyle-cochran, line 79: This function is defined in `Model`, so I think it needs the override, no?

kyle-cochran: Oh I see. The name is changed in Model...

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Exponential.cpp
```diff
@@ -1,16 +1,13 @@
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 /// @project        Open Space Toolkit ▸ Physics
-/// @file           OpenSpaceToolkit/Physics/Environment/Atmospheric/Exponential.cpp
+/// @file           OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Exponential.cpp
```
kyle-cochran, line 4: There's a similar situation in the Gravity modeling with the `Spherical` model. If we decide to move this one, I think we should move that one as well.

PR #100, opened 2023-05-06
fix: decouple testing commands from build-development-image command. …

kyle-cochran:
…Trust CI to do it.


Makefile
```diff
@@ -341,7 +338,7 @@ test-coverage: ## Run test coverage cpp
 
 	@ make test-coverage-cpp
 
-test-coverage-cpp: build-development-image
+test-coverage-cpp:
```
lucas-bremond, line 341: This will fail is the image hasn't been built no?

PR #90, opened 2022-11-09
Draft: feat: improve TOD frame definition

lucas-bremond:
This PR removes the bespoke TOD frame computation logic, and uses the SOFA library instead.

PR #109, opened 2023-05-30
feat: improve TOD frame definition

phc1990:
**Deltas w.r.t. [Lucas PR]**(https://github.com/open-space-collective/open-space-toolkit-physics/pull/90):

- Formatting (headers, footers, cpp clang, python black)
- `Frame.hpp`:
  - Removed `ICRF()` function as it was not implemented. 
- `Frame.cpp`:
  - Updated `TOD` to include IAU Theory in its name (e.g. `TOD (IAU 2006) @ 2000-01-01 12:00:00 [TT]`)
- `Frame.test.cpp`:
  - Include `J2000` in test cases where it was missing
  - Removed `ICRF` commented test lines and cases
  - Updated `TOD` cases to cope with IAU Theory
  - Updated `TOD` cases to assert that its parent frame is now `GCRF` (no longer `MOD`)
- `TOD.test.cpp` (in `Frame` folder):
  - Expanded "NoCaching" test scenario so that it considers not only different epochs same IAU Theory, but also same epoch different IAU Theories.


lucas-bremond:
And even easier-to-review approach would be to rebased this PR on top of https://github.com/open-space-collective/open-space-toolkit-physics/pull/90 to see the diffs.

PR #115, opened 2023-06-07
chore: update environment defaults to use newest spice files

kyle-cochran:


vishwa2710 (approved):
Is the goal just to get tests passing and unblock landing other MRs and then work on the improvement (regex?)

PR #121, opened 2023-06-13
feat: modelIsDefined functions in Celestial

kyle-cochran:
Add functions to check if gravity/magnetic/atmosphere models are defined for a celestial object. This allows for the model to be either a nullptr or set to Undefined (in the case of Earth)


bindings/python/test/environment/objects/test_celestial.py
```diff
@@ -0,0 +1,15 @@
+# Apache License 2.0
+
+from ostk.physics.environment.objects.celestial_bodies import Earth
+
+class TestCelestial:
+
+    def test_model_is_defined_success(self):
+        earth = Earth.spherical()
+    
+        assert earth.gravitational_model_is_defined() is True
```
Derollez, line 10:
```suggestion
        assert earth.gravitational_model_is_defined()
```

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Objects/Celestial.cpp
```diff
@@ -78,6 +78,9 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Objects_Celestial(pybind11::mo
         .def("get_position_in", &Celestial::getPositionIn)
         .def("get_transform_to", &Celestial::getTransformTo)
         .def("get_axes_in", &Celestial::getAxesIn)
+        .def("gravitational_model_is_defined", &Celestial::gravitationalModelIsDefined)
+        .def("magnetic_model_is_defined", &Celestial::magneticModelIsDefined)
+        .def("atmospheric_model_is_defined", &Celestial::atmosphericModelIsDefined)
```
Derollez, line 83: Move after the `access_*` functions to match the order from cpp code?

Derollez (approved):

bindings/python/test/environment/objects/test_celestial.py
kyle-cochran: I think we tend to be explicit in bool comparisons. See here for an example: https://github.com/open-space-collective/open-space-toolkit-physics/blob/main/bindings/python/test/coordinate/test_frame.py

PR #122, opened 2023-06-13
feat: (option 2) add high-level undefined types for environment models 

kyle-cochran:
that are not tied to specific planets.


lucas-bremond:
Please delete the branch from the remote if unnecessary!

PR #125, opened 2023-06-14
refactor: remove instant from object/celestial classes

vishwa2710:
* remove instant_ as a member variable for the Object parent class
* pass the instant as an argument for relevant functions
This allows the celestial bodies to remain immutable in the environment after creating them, while still allowing the environment class to keep store and pass instants as appropriate.

This will make it possible to get a shared pointer to a const Celestial in OSTk astro while instantiating the propagator from the environment.
Also it prevents us from having to clone celestials since we technically never need to modify them for the sake of updating an instant.

IMO the instant is something that should be supplied at the function call as it is not a configuration parameter for the object/celestial classes.

### DIFF
--- a/include/OpenSpaceToolkit/Physics/Environment/Objects/CelestialBodies/Moon.hpp
+++ b/include/OpenSpaceToolkit/Physics/Environment/Objects/CelestialBodies/Moon.hpp
@@ -54,8 +54,7 @@ class Moon : public Celestial
 
     Moon(
         const Shared<Ephemeris>& anEphemeris,
-        const MoonGravitationalModel::Type& aGravitationalModelType,
-        const Instant& anInstant
+        const MoonGravitationalModel::Type& aGravitationalModelType
     );
 
     /// @brief              Destructor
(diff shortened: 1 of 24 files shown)


include/OpenSpaceToolkit/Physics/Environment/Objects/CelestialBodies/Moon.hpp
```diff
@@ -52,10 +52,7 @@ class Moon : public Celestial
     /// (Spherical model only)
     /// @param              [in] anInstant An instant
 
-    Moon(
-        const Shared<Ephemeris>& anEphemeris,
-        const MoonGravitationalModel::Type& aGravitationalModelType
-    );
+    Moon(const Shared<Ephemeris>& anEphemeris, const MoonGravitationalModel::Type& aGravitationalModelType);
```
lucas-bremond, line 55: Odd that the formatter is touching this, no?

lucas-bremond:
This PR is a draft right? (I didn't see any diff related to the intent)

You may want to prefix the title with `Draft:` to prevent review :)

vishwa2710:
It's ready for review!

Derollez (approved):
Neat!

PR #127, opened 2023-06-15
build: make shared object version fetching generic

vishwa2710:
Not sure if this is the best way to move forward but now it looks for libopen-space-toolkit-physics.so.1: https://github.com/open-space-collective/open-space-toolkit-physics/actions/runs/5283347064/jobs/9560019478


tools/development/start.sh
```diff
@@ -58,7 +58,7 @@ if [[ ! -z ${1} ]] && [[ ${1} == "--link" ]]; then
             cp -as /mnt/open-space-toolkit-core/include/OpenSpaceToolkit/Core /usr/local/include/OpenSpaceToolkit/Core; \
             cp -as /mnt/open-space-toolkit-core/src/OpenSpaceToolkit/Core/* /usr/local/include/OpenSpaceToolkit/Core/; \
             ln -s /mnt/open-space-toolkit-core/lib/libopen-space-toolkit-core.so /usr/local/lib/; \
-            ln -s /mnt/open-space-toolkit-core/lib/libopen-space-toolkit-core.so.0 /usr/local/lib/;"
+            ln -s /mnt/open-space-toolkit-core/lib/libopen-space-toolkit-core.so.1 /usr/local/lib/;"
```
Derollez, line 61: So these are the shared object "soname" of the dependencies of this project. I don't think we need to touch them right now. That being said we need to make this more generic across the repo (probably using the `.env` approach to have dependencies versions available in the Makefile)

Derollez (commented):
Adding this PR on top to align and make it a bit more generic temporarily. Will add another one on top to leverage a `versions.env` file for dependencies to make it even more generic across the repository

tools/testing/python/python-3.6/run.sh
```diff
@@ -74,7 +74,7 @@ run_container ()
     --rm \
```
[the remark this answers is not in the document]
Derollez: Applied in suggestion PR

tools/development/start.sh
Derollez: Reverted for now to so.0 and will propose an additional PR with `versions.env` file

Derollez (approved):

PR #137, opened 2023-06-21
feat: add manager for CSSI space weather data

kyle-cochran:


src/OpenSpaceToolkit/Physics/Environment/Ephemerides/SPICE/Engine.cpp
```diff
@@ -234,7 +234,7 @@ Array<Kernel> Engine::DefaultKernels(const Directory& aLocalRepository)
 
     static const Array<Kernel> defaultKernels = {
 
-        Manager::Get().findKernel("naif[0-9]*.tls"),                                      // Leap seconds
+        Manager::Get().findKernel("naif[0-9]*\\.tls"),                                    // Leap seconds
```
kyle-cochran, line 237: tiny bug in kernel fetching. Not breaking

docker/development/Dockerfile
```diff
@@ -106,7 +106,7 @@ RUN git clone https://github.com/geographiclib/geographiclib /tmp/geographiclib
 
 ## Open Space Toolkit ▸ Core
 
-ARG OSTK_CORE_VERSION="0.6.0"
+ARG OSTK_CORE_VERSION="0.7.0"
```
Derollez, line 109: All good with using this (it will need get propagated to downstream repos ⚠️). Also let's bump `requirements.py` to have them in sync if all tested

kyle-cochran: Hmmm... good point. Wasn't this was the thing that was breaking Requesting pipeline? This is not actually needed, I just wanted the stack traces for development, so if it will break I can revert to 0.6

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.cpp
```diff
@@ -0,0 +1,588 @@
...
+    const auto predictionIt = dailyPredictions_.find(instantMjd.floor());
+
+    if (predictionIt != dailyPredictions_.end())
+    {
+        return predictionIt->second;
+    }
+    else
+    {
+        throw ostk::core::error::RuntimeError("Cannot find prediction at [{}].", anInstant.toString(Scale::UTC));
+    }
```
vishwa2710, line 284:
```suggestion
    throw ostk::core::error::RuntimeError("Cannot find prediction at [{}].", anInstant.toString(Scale::UTC));
```
else is redundant

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.cpp
```diff
@@ -0,0 +1,588 @@
...
+    const auto predictionIt = monthlyPredictions_.find(monthMjd.floor());
+
+    if (predictionIt != monthlyPredictions_.end())
+    {
+        return predictionIt->second;
+    }
+    else
+    {
+        throw ostk::core::error::RuntimeError("Cannot find prediction at [{}].", anInstant.toString(Scale::UTC));
+    }
```
vishwa2710, line 333:
```suggestion
    throw ostk::core::error::RuntimeError("Cannot find prediction at [{}].", anInstant.toString(Scale::UTC));
```
else is redundant

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.cpp
```diff
@@ -0,0 +1,588 @@
...
+}
+
+CSSISpaceWeather CSSISpaceWeather::Undefined()
+{
+    return CSSISpaceWeather();
+}
+
+CSSISpaceWeather CSSISpaceWeather::Load(const fs::File& aFile)
```
vishwa2710, line 341:
```suggestion
CSSISpaceWeather CSSISpaceWeather::Load(const File& aFile)
```

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.cpp
```diff
@@ -0,0 +1,588 @@
...
+{
+    using ostk::core::types::Index;
+    using ostk::core::types::Uint8;
+    using ostk::core::types::Uint16;
+    using ostk::core::types::Real;
+    using ostk::core::types::String;
+
+    using ostk::physics::time::Scale;
+    using ostk::physics::time::Time;
+    using ostk::physics::time::DateTime;
```
vishwa2710, line 351: Move to the top of the file

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.cpp
```diff
@@ -0,0 +1,588 @@
...
+    {
+        Array<String> results;
+
+        if (s.empty())
+        {
+            return results;
+        }
+
+        size_t last = 0;
+        size_t next = 0;
```
vishwa2710, line 378:
```suggestion
        Size last = 0;
        Size next = 0;
```
add `using ostk::core::types::Size` on top

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.cpp
```diff
@@ -0,0 +1,588 @@
...
+    {
+        throw ostk::core::error::RuntimeError("File [{}] does not exist.", aFile.toString());
+    }
+
+    CSSISpaceWeather spaceWeather;
+    std::ifstream fileStream {aFile.getPath().toString()};
+
+    String line;
+
+    auto split = [](String s, const String delimiter) -> Array<String>
```
vishwa2710, line 368:
```suggestion
    TBI: Use split from string type once new OSTk Core released
    auto split = [](String s, const String delimiter) -> Array<String>
```

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.cpp
```diff
@@ -0,0 +1,588 @@
...
+        {
+            results.add(s.substr(last, next - last));
+            last = next + 1;
+        }
+
+        results.add(s.substr(last));
+        return results;
+    };
+
+    while (std::getline(fileStream, line))
```
vishwa2710, line 389: Could you use `Table::Load(aFile, Table::Format::CSV, true)` and then maybe loop through the rows?

vishwa2710: from ostk::core::ctnr::Table?

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.cpp
```diff
@@ -0,0 +1,588 @@
...
+            Instant::ModifiedJulianDate(Real::Integer(spaceWeather.monthlyPredictions_.rbegin()->first), Scale::UTC);
+
+        spaceWeather.monthlyPredictionInterval_ =
+            Interval::Closed(monthlyPredictionStartInstant, monthlyPredictionEndInstant);
+    }
+
+    return spaceWeather;
+}
+
+CSSISpaceWeather::CSSISpaceWeather()
```
vishwa2710, line 570: Why have a private default constructor at all? And not just use `CSSISpaceWeather::Undefined()` and define this over there?

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
```diff
@@ -0,0 +1,790 @@
...
+#include <OpenSpaceToolkit/IO/IP/TCP/HTTP/Client.hpp>
+
+#include <OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.hpp>
+#include <OpenSpaceToolkit/Physics/Time/Date.hpp>
+#include <OpenSpaceToolkit/Physics/Time/DateTime.hpp>
+#include <OpenSpaceToolkit/Physics/Time/Instant.hpp>
+#include <OpenSpaceToolkit/Physics/Time/Scale.hpp>
+#include <OpenSpaceToolkit/Physics/Time/Time.hpp>
+
+#include <experimental/filesystem>
```
vishwa2710, line 25: What is this? O_o

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
```diff
@@ -0,0 +1,790 @@
...
+Directory Manager::getLocalRepository() const
+{
+    std::lock_guard<std::mutex> lock {mutex_};
+
+    return localRepository_;
+}
+
+Directory Manager::getCSSISpaceWeatherDirectory() const
+{
+    using ostk::core::fs::Path;
```
vishwa2710, line 70: Move to top

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
```diff
@@ -0,0 +1,790 @@
...
+
+            return Array<Integer> {
+                observation.KP1,
+                observation.KP2,
+                observation.KP3,
+                observation.KP4,
+                observation.KP5,
+                observation.KP6,
+                observation.KP7,
+                observation.KP8};
```
vishwa2710, line 133:
```suggestion
                observation.KP8,
            };
```

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
```diff
@@ -0,0 +1,790 @@
...
+
+            return Array<Integer> {
+                prediction.KP1,
+                prediction.KP2,
+                prediction.KP3,
+                prediction.KP4,
+                prediction.KP5,
+                prediction.KP6,
+                prediction.KP7,
+                prediction.KP8};
```
vishwa2710, line 147:
```suggestion
                prediction.KP8,
            };
```

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
```diff
@@ -0,0 +1,790 @@
...
+            return Array<Integer> {
+                observation.AP1,
+                observation.AP2,
+                observation.AP3,
+                observation.AP4,
+                observation.AP5,
+                observation.AP6,
+                observation.AP7,
+                observation.AP8,
+                observation.AP_AVG};
```
vishwa2710, line 182:
```suggestion
                observation.AP_AVG,
            };
```

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
```diff
@@ -0,0 +1,790 @@
...
+            return Array<Integer> {
+                prediction.AP1,
+                prediction.AP2,
+                prediction.AP3,
+                prediction.AP4,
+                prediction.AP5,
+                prediction.AP6,
+                prediction.AP7,
+                prediction.AP8,
+                prediction.AP_AVG};
```
vishwa2710, line 197:
```suggestion
                prediction.AP_AVG,
            };
```

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
```diff
@@ -0,0 +1,790 @@
...
+    {
+        return CSSISpaceWeatherMap.rbegin()->second;
+    }
+
+    return const_cast<Manager*>(this)->fetchLatestCSSISpaceWeather_();
+}
+
+void Manager::setup()
+{
+    using ostk::core::fs::Path;
```
vishwa2710, line 573: Move to top

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
```diff
@@ -0,0 +1,790 @@
...
+    using ostk::core::ctnr::Map;
+    using ostk::core::fs::Path;
+
+    using ostk::io::ip::tcp::http::Client;
+
+    using ostk::physics::time::Scale;
+    using ostk::physics::time::Date;
+    using ostk::physics::time::Time;
+    using ostk::physics::time::DateTime;
+    using ostk::physics::time::Instant;
```
vishwa2710, line 617: Move to top

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
```diff
@@ -0,0 +1,790 @@
...
+            throw ostk::core::error::RuntimeError("CSSI Space Weather already added.");
+        }
+    }
+
+    CSSISpaceWeatherArray_.add(aCSSISpaceWeather);  // [TBI] Add in ascending time order
+
+    CSSISpaceWeatherIndex_ = 0;
+}
+
+File Manager::fetchLatestCSSISpaceWeather_()
```
vishwa2710, line 602: Why underscore in function name?

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
```diff
@@ -0,0 +1,790 @@
...
+
+    Directory temporaryDirectory =
+        Directory::Path(this->getCSSISpaceWeatherDirectory().getPath() + Path::Parse(temporaryDirectoryName));
+
+    this->lockLocalRepository(localRepositoryLockTimeout_);
+
+    const URL latestCSSISpaceWeatherUrl = remoteUrl_ + CSSISpaceWeatherFileName;
+    std::cout << latestCSSISpaceWeatherUrl.toString() << std::endl;
+    File latestCSSISpaceWeatherFile = File::Undefined();
+    Directory destinationDirectory = Directory::Undefined();
```
vishwa2710, line 629:
```suggestion
    const URL latestCSSISpaceWeatherUrl = remoteUrl_ + CSSISpaceWeatherFileName;
    std::cout << latestCSSISpaceWeatherUrl.toString() << std::endl;

    File latestCSSISpaceWeatherFile = File::Undefined();
    Directory destinationDirectory = Directory::Undefined();
```
Nit: spacing
Also maybe add a header to the std::cout above for some more context?

vishwa2710 (commented):
First pass, will do another tomorrow on the tests.

test/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.test.cpp
```diff
@@ -209,7 +209,7 @@ TEST(OpenSpaceToolkit_Physics_Environment_Atmospheric_Earth, GetDensityAt_LLA)
                 }
                 catch (const RuntimeError& e)
                 {
-                    EXPECT_EQ(expectedString, e.what());
+                    EXPECT_EQ(expectedString, e.getMessage());
```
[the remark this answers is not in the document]
kyle-cochran: Yes it is :). Will revert this when reverting core version

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.hpp
```diff
@@ -0,0 +1,172 @@
...
+    static CSSISpaceWeather Load(const fs::File& aFile);
+
+   private:
+    Date lastObservationDate_;
+
+    Interval observationInterval_;
+    Map<Integer, CSSISpaceWeather::Observation> observations_;
+
+    Interval dailyPredictionInterval_;
+    Map<Integer, CSSISpaceWeather::DailyPrediction> dailyPredictions_;
```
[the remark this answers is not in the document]
kyle-cochran: It's the same data format, but different data. I think it's worth keeping it separate to make it clear when we're using prediction and to be able to prefer observations over predictions.

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
kyle-cochran: These functions are private in the header

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.cpp
```diff
@@ -0,0 +1,588 @@
...
+
+        if (lineParts.empty() || lineParts[0] == "DATE")
+        {
+            continue;
+        }
+
+        // Data points that exist for every reading
+        const Date date = Date::Parse(lineParts[0], Date::Format::Standard);
+
+        // [TBR] Toss data past 2030 due to this restriction in the Instant class
```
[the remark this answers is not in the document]
kyle-cochran: Yeah, though I'm not sure why. Meant to ping about this. It's here in the Instant class:
https://github.com/open-space-collective/open-space-toolkit-physics/blob/main/src/OpenSpaceToolkit/Physics/Time/Instant.cpp#L430

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.hpp
```diff
@@ -0,0 +1,172 @@
...
+namespace environment
+{
+namespace atmospheric
+{
+namespace earth
+{
+namespace weather
+{
+
+namespace fs = ostk::core::fs;
```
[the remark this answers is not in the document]
kyle-cochran: Good catch. This is leftover from some intermediate code 🙏

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.hpp
```diff
@@ -0,0 +1,172 @@
...
+///                             which are commonly used to model atmospheric density.
+///
+/// @ref                        http://celestrak.org/SpaceData/SpaceWx-format.php
+
+class CSSISpaceWeather
+{
+   public:
+    struct Observation
+    {
+        Date date;     /// UTC day of reading.
```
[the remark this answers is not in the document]
kyle-cochran: Yeah let me try aligning and see if clang is okay with it

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.hpp
```diff
@@ -0,0 +1,172 @@
...
+        Integer KP_SUM;  /// Sum of the 8 Kp indices for the day.
+        Integer AP1;     /// Planetary Equivalent Amplitude (Ap) for 0000-0300 UT.
+        Integer AP2;     /// Planetary Equivalent Amplitude (Ap) for 0300-0600 UT.
+        Integer AP3;     /// Planetary Equivalent Amplitude (Ap) for 0600-0900 UT.
+        Integer AP4;     /// Planetary Equivalent Amplitude (Ap) for 0900-1200 UT.
+        Integer AP5;     /// Planetary Equivalent Amplitude (Ap) for 1200-1500 UT.
+        Integer AP6;     /// Planetary Equivalent Amplitude (Ap) for 1500-1800 UT.
+        Integer AP7;     /// Planetary Equivalent Amplitude (Ap) for 1800-2100 UT.
+        Integer AP8;     /// Planetary Equivalent Amplitude (Ap) for 2100-0000 UT.
+        Integer AP_AVG;  /// Arithmetic average of the 8 Ap indices for the day.
```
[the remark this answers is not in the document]
kyle-cochran: Yeah the idea was to match the CSV, which uses all caps snake case. (@Derollez that link is for the legacy format, this uses the CSV format described [here](https://celestrak.org/SpaceData/SpaceWx-format.php)). 

Honestly, I meant to revisit this but it slipped. I think CamelCase as Remy pointed out looks a lot cleaner (especially to indicate the subscripts). Will update!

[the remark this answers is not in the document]
kyle-cochran: Alright clang format doesn't allow everything to be fully lined up, but since renaming the variable, the comments are _mostly_ lined up. (although add a `///<` makes the alignment very wonky. Current version looks pretty clean. Think it's good enough?

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
```diff
@@ -0,0 +1,790 @@
...
+        else if (CSSISpaceWeatherPtr->accessMonthlyPredictionInterval().contains(dayInstant))
+        {
+            return CSSISpaceWeatherPtr->accessMonthlyPredictionAt(dayInstant).F107_OBS;
+        }
+    }
+
+    throw ostk::core::error::RuntimeError("Cannot obtain F10.7 Solar Flux at [{}].", anInstant.toString());
+}
+
+Real Manager::getF107SolarFlux81DayAvgAt(const Instant& anInstant) const
```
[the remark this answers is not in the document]
kyle-cochran: Totally preference here, but 81-day Average feels slightly more natural to me than 81-days Average. If someone feels strongly I can change it :)

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.hpp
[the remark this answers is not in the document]
kyle-cochran: It was not 🙃 (mostly)

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth.cpp
```diff
@@ -53,4 +55,12 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth(pybind11::mo
...
             .value("Exponential", Earth::Type::Exponential);
     }
+
+    // Create "earth" python submodule
+    auto earth = aModule.def_submodule("earth");
+
+    // Add __path__ attribute for "earth" submodule
+    earth.attr("__path__") = "ostk.physics.environment.atmospheric.earth";
+
+    OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth_Weather(earth);  // TBI
```
vishwa2710, line 65: Is this still `TBI`?

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth/Weather.cpp
```diff
@@ -0,0 +1,17 @@
+/// Apache License 2.0
+
+#include <OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.cpp>
+#include <OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth/Weather/Manager.cpp>
+
+inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth_Weather(pybind11::module& aModule)
```
vishwa2710, line 6: Nit:
```suggestion
inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth_SpaceWeather(pybind11::module& aModule)
```
Maybe weather is a bit too generic?
If you agree then we can update the package path to `space_weather`

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.hpp
```diff
@@ -0,0 +1,170 @@
...
+    Interval observationInterval_;
+    Map<Integer, CSSISpaceWeather::Observation> observations_;
+
+    Interval dailyPredictionInterval_;
+    Map<Integer, CSSISpaceWeather::DailyPrediction> dailyPredictions_;
+
+    Interval monthlyPredictionInterval_;
+    Map<Integer, CSSISpaceWeather::MonthlyPrediction> monthlyPredictions_;
+
+    CSSISpaceWeather();
```
vishwa2710, line 160: Sorry I didn't make this clear before but what is the point of having a private default constructor vs just the undefined static constructor?

vishwa2710 (commented):
Few more nits.

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.hpp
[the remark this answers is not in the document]
Derollez: Ah my bad, misread totally the comments

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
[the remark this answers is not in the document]
Derollez: Nope, I don't speak English properly, let's keep it this way of more natural ✅

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
```diff
@@ -0,0 +1,790 @@
...
+using ostk::core::ctnr::Map;
+using ostk::core::fs::Path;
+
+using ostk::physics::time::Scale;
+using ostk::physics::time::Date;
+using ostk::physics::time::Time;
+using ostk::physics::time::DateTime;
+using ostk::physics::time::Instant;
+
+const String CSSISpaceWeatherFileName = "SW-Last5Years.csv";
```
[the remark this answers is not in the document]
Derollez: OK!

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth.cpp
Derollez: nope let's rm

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth/Weather.cpp
Derollez: then we need to keep it in sync with cpp code

kyle-cochran: I think the difference between `Weather` and `SpaceWeather` is pretty minimal in the context of OSTk.

Now that I'm looking at it though, I think we can completely remove this subdirectory. It's already under `Atmospheric`, so weather is a bit redundant. What's more is that it deviates from what we do for some of the other things with Managers.

Gonna just nix this directory. Thanks for the nit :)

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.hpp
kyle-cochran: I see your point. This is a carry-over from the one I copied from, but I don't think we need it. Let me try it.

kyle-cochran: Looks like prior to C++ 11, this was the preferred way to hide the default constructor. Probably a holdover from that. [ref](https://stackoverflow.com/questions/10474417/how-to-delete-the-default-constructor)

kyle-cochran: Ah wait, I think you actually _do_ need both. 

If you remove the empty Constructor, there are no constructors left to call. the `CSSISpaceWeather::Undefined()` function has to call _something_, so this is needed.

You could alternatively get rid of the `Undefined()` function and just keep the default empty constructor, but I think that makes it a bit less clear that it is not defined.

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
vishwa2710: Interesting gotcha. I didn't realize we use an underscore in function naming as well. Let me take a look.

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.hpp
vishwa2710: Sorry I meant that you remove the private constructor, and just have the Undefined static constructor. And you move the logic from the private constructor to the Static undefined constructor. This is because for most of OSTk we don't allow a default constructor. In this case it's private which is 🧷 but it's not used anywhere so it's kinda pointless (unless I'm missing something).

kyle-cochran: Maybe I'm not understanding the pattern that you mean. From what I understand, `Undefined()` is not a constructor, but a static factory function that calls a constructor. 

In other cases that I've seen, we always have at least one non-default constructor, then `Class::Undefined()` can call that with undefined arguments.
In this case if we remove the default constructor, there are no non-default constructor so `Class::Undefined()` has nothing to call and it can't make an object.

Very possible this is a gap in my cpp-fu 🙏

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.cpp
kyle-cochran: Yeah I realized that the `Table` class existed only after writing this 😅. Totally can do that. I'm not sure how much it buys us? I don't know if it would reduce the code needed here by much. Thoughts?

vishwa2710: I think it helps abstract the File I/O and also it removes the need for `lexical_cast` as it converts objects to String, Real and Int appropriately. I would say it's worth it as it might reduce a fair bit of code but I could be wrong. I'm happy to open a suggestion MR.

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
vishwa2710: Ah gotcha it's cause there is another function with the same name, so we have a private one with an _. Gotcha.

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/Manager.cpp
kyle-cochran: It's to check file size:
`std::experimental::filesystem::file_size`

but looks like it's no longer experimental! Changed to `#include <filesystem>`

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.hpp
vishwa2710: Ah yeah fair point that there is no other constructor, I missed that sorry. I'm going to push a suggestion commit on top to match a better pattern in my opnion.

kyle-cochran: Sounds good, although I would say that if we're changing the pattern, we should probably do it in all of the Managers so that it remains a pattern, no? 

What kinda pattern you thinking? I'm happy to help with that!

kyle-cochran: We hashed it out in this thread https://loftorbital.slack.com/archives/C03GCNLG8G5/p1687393460392469

Decided that the CSSISpaceWeather class might not be the best place to inject test data. Got offline :blessing: from Vishwa

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth/Manager.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    class_<Manager> manager(aModule, "Manager");
+
+    manager
+
+        .def("get_mode", &Manager::getMode)
+        .def("get_local_repository", &Manager::getLocalRepository)
+        .def("get_cssi_space_weather_directory", &Manager::getCSSISpaceWeatherDirectory)
+        .def("get_remote_url", &Manager::getRemoteUrl)
+        .def("get_cssi_space_weather_array", &Manager::getCSSISpaceWeatherArray)  // TBI: check case
```
vishwa2710, line 21: What is this TBI?

bindings/python/test/environment/atmospheric/earth/test_manager.py
```diff
@@ -0,0 +1,204 @@
...
+
+    manager.set_mode(Manager.Mode.Automatic)
+    manager.set_remote_url(
+        URL.parse("https://celestrak.org/SpaceData/")
+    )  # SW-Last5Years.csv
+
+    yield manager
+
+    manager.reset()
+    manager.clear_local_repository()
```
vishwa2710, line 35: Move to the conftest.py?

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/CSSISpaceWeather.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+    const Reading& accessDailyPredictionAt(const Instant& anInstant) const;
+
+    const Interval& accessMonthlyPredictionInterval() const;
+
+    const Reading& accessMonthlyPredictionAt(const Instant& anInstant) const;
+
+    static CSSISpaceWeather Undefined();
+
+    static CSSISpaceWeather Load(const File& aFile);
```
vishwa2710, line 118: Can we add docstrings for these?

vishwa2710 (approved):
v nit.

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/CSSISpaceWeather.hpp
Derollez: yeah agreed

bindings/python/test/environment/atmospheric/earth/test_manager.py
vishwa2710: Fair point, as long as it's in earth it should be okay?

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth/Manager.cpp
vishwa2710: Oh gotcha yeah. Thanks that makes sense I agree with lower case CSSI.

bindings/python/test/environment/atmospheric/earth/test_manager.py
kyle-cochran: Good idea

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth/Manager.cpp
kyle-cochran: good catch

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/CSSISpaceWeather.hpp
kyle-cochran: Me & github copilot are on it

kyle-cochran: Added docstrings to the reference files Bulletin A/B + Finals for good measure

Derollez: Thanks a lot for the alignment effort 🙏

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Weather/CSSISpaceWeather.cpp
kyle-cochran: Hmm if it removes the lexical cast, I am all for it 🙏. Good suggestion, I'll try it.

vishwa2710 (approved):
Fantabulous

PR #139, opened 2023-06-23
feat: suggestions on atmospheric weather

vishwa2710:


kyle-cochran:
We hashed it out and decided that this class might not be most convenient place to inject test data and so decided to leave the constructor situation as-is for now.

PR #143, opened 2023-07-05
fix: use https instead of ftp to get SPICE kernels

Derollez:


test/OpenSpaceToolkit/Physics/Environment/Ephemerides/SPICE.test.cpp
```diff
@@ -63,7 +63,7 @@ TEST(OpenSpaceToolkit_Physics_Environment_Ephemerides_SPICE, IsDefined)
     }
 }
 
-TEST(OpenSpaceToolkit_Physics_Environment_Ephemerides_SPICE, AccessFrame)
+TEST(OpenSpaceToolkit_Physics_Environment_Ephemerides_SPICE, e)
```
Derollez, line 66:
```suggestion
TEST(OpenSpaceToolkit_Physics_Environment_Ephemerides_SPICE, AccessFrame)
```

Derollez:
Amazing @kyle-cochran !

src/OpenSpaceToolkit/Physics/Environment/Ephemerides/SPICE/Engine.cpp
```diff
@@ -13,7 +13,7 @@
 #include <OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/Dynamic.hpp>
 
 // clang-format off
-// Include Manager.hpp before Engine.hpp to avoid name conflict for "ostk::core::types::Index" 
+// Include Manager.hpp before Engine.hpp to avoid name conflict for "ostk::core::types::Index"
 #include <OpenSpaceToolkit/Physics/Environment/Ephemerides/SPICE/Manager.hpp>
 #include <OpenSpaceToolkit/Physics/Environment/Ephemerides/SPICE/Engine.hpp>
```
vishwa2710, line 18: I believe instead of this, we can merely use an alias in Engine. IMO that's a better solution as there is an actual conflict here so it's safer to use an alias like
```
using IndexType = ostk::core::types::Index
```

PR #146, opened 2023-07-12
refactor: move eclipse method from utils to environment

vishwa2710:
I believe it's better suited there, and thus also easily exposed via python bindings.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment.cpp
+++ b/bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment.cpp
@@ -49,6 +49,8 @@ inline void OpenSpaceToolkitPhysicsPy_Environment(pybind11::module& aModule)
 
         .def("set_instant", &Environment::setInstant, arg("instant"))
 
+        .def("is_position_in_eclipse", &Environment::isPositionInEclipse, arg("position"))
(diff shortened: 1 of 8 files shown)


vishwa2710:
1. Technically not a breaking change as I don't believe anything is using this function (yet). Also there are no bindings for it so it doesn't affect any python code.
2. So the reason it's a part of the environment is because we can calculate eclipse with all the objects in the environment (not just the earth), and leverage the presence of the sun which is already part of the environment.

We could definitely do it the way you suggested but that interface only allows the computation of an eclipse with a single body. Also, in general the OSTk philosophy is to rely on the environment when convenient. However, I do agree that having the flexibility to do it with celestials would be nice... Let me see if I can make both work.

kyle-cochran:
Ah I actually didn't realize that it compared _everything_ in the environment. That makes sense to me then.

The only other use case that I can think of is if we wanted to use a different light source like the moon. (to check secondary start tracker effects). But I agree that for secondary cases like that we can do the intersection directly. 
We could also probably add the light source as an optional argument in the future in a non-breaking way if we ever need that flexibility.

lgtm

kyle-cochran (approved):

Derollez (approved):

PR #151, opened 2023-07-26
chore: align Angle python bindings with ostk mathematics and add tests

AlexandreLoftOrbital:


bindings/python/test/units/derived/test_angle.py
```diff
@@ -15,3 +19,337 @@ def test_angle_conversion():
...
+    assert angle.in_degrees() == 270.0
+
+    angle /= 3.0
+    assert angle is not None
+    assert isinstance(angle, AnglePhysics)
+    assert angle.get_unit() == Unit.Degree
+    assert angle.in_degrees() == 90.0
+
+
+# def test_angle_between_vector2d ():
```
phc1990, line 269: Should these 3 comments be removed?

Derollez: This is none implemented pieces of code. Let's clean and remove I'd say

Derollez (approved):

bindings/python/test/units/derived/test_angle.py
AlexandreLoftOrbital: done

bindings/python/src/OpenSpaceToolkitPhysicsPy/Units/Derived/Angle.cpp
```diff
@@ -77,70 +78,69 @@ inline void OpenSpaceToolkitPhysicsPy_Units_Derived_Angle(pybind11::module& aMod
...
-            {
-                return Angle::Between(aFirstVector, aSecondVector);
-            }
+        .def_static(
+            "between",
+            overload_cast<const Vector3d&, const Vector3d&>(&Angle::Between),
+            arg("first_vector"),
+            arg("second_vector")
         )
-        .def_static("parse", &Angle::Parse)
```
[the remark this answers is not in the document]
Derollez: let's keep them here if the cpp functions are in the code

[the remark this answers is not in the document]
AlexandreLoftOrbital: added the binding back and added the appropriate test

bindings/python/src/OpenSpaceToolkitPhysicsPy/Units/Derived/Angle.cpp
```diff
@@ -77,70 +78,69 @@ inline void OpenSpaceToolkitPhysicsPy_Units_Derived_Angle(pybind11::module& aMod
...
-            "between_vector3d",
-            +[](const Vector3d& aFirstVector, const Vector3d& aSecondVector) -> Angle
-            {
-                return Angle::Between(aFirstVector, aSecondVector);
-            }
+        .def_static(
+            "between",
+            overload_cast<const Vector3d&, const Vector3d&>(&Angle::Between),
+            arg("first_vector"),
+            arg("second_vector")
```
Derollez, line 134: agreed that this is cleaner to have overload in python reflect the overload in cpp but this is breaking potentially in some notebooks, just fyi to update the rest accordingly in parallel

bindings/python/src/OpenSpaceToolkitPhysicsPy/Units/Derived/Angle.cpp
```diff
@@ -77,70 +78,69 @@ inline void OpenSpaceToolkitPhysicsPy_Units_Derived_Angle(pybind11::module& aMod
...
-            {
-                return anAngle.inArcseconds();
-            }
+            overload_cast<const Real&, const Real&>(&Angle::inArcseconds, const_),
+            arg("lower_bound"),
+            arg("upper_bound")
         )
         .def("in_revolutions", &Angle::inRevolutions)
-        .def("to_string", &Angle::toString, "aPrecision"_a = Integer::Undefined())
+        .def("to_string", &Angle::toString, arg("do_sanitize") = false)
```
Derollez, line 111: can we keep the default value for `aPrecision`?

bindings/python/src/OpenSpaceToolkitPhysicsPy/Units/Derived/Angle.cpp
```diff
@@ -108,7 +109,7 @@ inline void OpenSpaceToolkitPhysicsPy_Units_Derived_Angle(pybind11::module& aMod
             arg("upper_bound")
         )
         .def("in_revolutions", &Angle::inRevolutions)
-        .def("to_string", &Angle::toString, arg("do_sanitize") = false)
+        .def("to_string", &Angle::toString, "aPrecision"_a = Integer::Undefined())
```
lucas-bremond, line 112: use `arg` form instead to be in sync w/ other bindings, and `precision` instead of `aPrecision`

AlexandreLoftOrbital: Updated
However when doing so, i noticed that the arg name is wrong in the current bindings
I changed it from `aPrecision` to `precision`, but i think it will be difficult to detect where it's breaking code in other project because `to_string` is a very common name
Also added a unit test for that
What is your thinking @lucas-bremond @Derollez

vishwa2710: I think that should be okay, not a lot of notebooks use the toString representation for Angle, and the default precision is often used. I think you should be good to go!

bindings/python/src/OpenSpaceToolkitPhysicsPy/Units/Derived/Angle.cpp
lucas-bremond: Nice!!

bindings/python/test/units/derived/test_angle.py
```diff
@@ -2,9 +2,13 @@
 
 import pytest
 
+import math
+
```
lucas-bremond, line 6:
```suggestion
import math

import pytest

```

bindings/python/src/OpenSpaceToolkitPhysicsPy/Units/Derived/Angle.cpp
AlexandreLoftOrbital: i checked and we should be good with the change

lucas-bremond (approved):

PR #155, opened 2023-08-04
fix: gap in bulletinA intervals

kyle-cochran:
The bulletin A file defines about a week's worth of observations for reference frame data. It then defines months of predictive data. So we have:
```
   23  7 21  60146 0.23587 .00009 0.49114 .00009 -0.020431 0.000012          
   [other observations]...       
   23  7 27  60152 0.24985 .00009 0.48239 .00009 -0.018731 0.000014 
  ---------
   2023  7 28  60153       0.2522      0.4808     -0.01819         
   2023  7 29  60154       0.2545      0.4792     -0.01744
   [other predictions]...
   2024  7 26  60517       0.1872      0.4533      0.00902
```

When we load the Bulletin A files, we record the interval associated with each. Which is:
```
[2023-07-21 00:00:00 - 2023-07-27 00:00:00]
[2023-07-28 00:00:00 - 2024-07-26 00:00:00]
```

This causes a 1-day gap in Bulletin A data. A point at `2023-07-27 12:00:00` is neither in the observation nor prediction interval.

I believe this is why the test in Physics that tries to grab data from 8 days ago can periodically fail. It does so whenever we are 8 days past that gap but IERS has not yet released a new bulletin.

This MR fixes the gap by:
- Changing the observation Interval to be open on the right side, and extending it by 1 day such that it capture all points in the last day.
- Using linear extrapolation for points that land in the gap between observation and predictions.

### DIFF
--- a/bindings/python/test/coordinate/frame/providers/iers/test_bulletin_a.py
+++ b/bindings/python/test/coordinate/frame/providers/iers/test_bulletin_a.py
@@ -36,7 +36,7 @@ class TestBulletinA:
     def test_access_observation_interval_success(self, bulletin_a: BulletinA):
         assert (
             bulletin_a.access_observation_interval().to_string()
-            == "[2020-10-23 00:00:00 - 2020-10-29 00:00:00] [UTC]"
+            == "[2020-10-23 00:00:00 - 2020-10-30 00:00:00] [UTC]"
         )
 
     def test_access_prediction_interval_success(self, bulletin_a: BulletinA):
@@ -59,7 +59,7 @@ class TestBulletinA:
     def test_get_observation_interval_success(self, bulletin_a: BulletinA):
         assert (
             bulletin_a.get_observation_interval().to_string()
-            == "[2020-10-23 00:00:00 - 2020-10-29 00:00:00] [UTC]"
+            == "[2020-10-23 00:00:00 - 2020-10-30 00:00:00] [UTC]"
         )
 
     def test_get_observation_at_success(self, bulletin_a: BulletinA):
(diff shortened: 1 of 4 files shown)


bindings/python/test/coordinate/frame/providers/iers/test_bulletin_a.py
```diff
@@ -36,7 +36,7 @@ def test_access_tai_minus_utc_epoch_success(self, bulletin_a: BulletinA):
     def test_access_observation_interval_success(self, bulletin_a: BulletinA):
         assert (
             bulletin_a.access_observation_interval().to_string()
-            == "[2020-10-23 00:00:00 - 2020-10-29 00:00:00] [UTC]"
+            == "[2020-10-23 00:00:00 - 2020-10-30 00:00:00] [UTC]"
```
vishwa2710, line 39:
```suggestion
            == "[2020-10-23 00:00:00 - 2020-10-30 00:00:00[ [UTC]"
```

bindings/python/test/coordinate/frame/providers/iers/test_bulletin_a.py
```diff
@@ -59,7 +59,7 @@ def test_get_tai_minus_utc_epoch_success(self, bulletin_a: BulletinA):
     def test_get_observation_interval_success(self, bulletin_a: BulletinA):
         assert (
             bulletin_a.get_observation_interval().to_string()
-            == "[2020-10-23 00:00:00 - 2020-10-29 00:00:00] [UTC]"
+            == "[2020-10-23 00:00:00 - 2020-10-30 00:00:00] [UTC]"
```
vishwa2710, line 62:
```suggestion
            == "[2020-10-23 00:00:00 - 2020-10-30 00:00:00[ [UTC]"
```

vishwa2710 (approved):
Looks good man!
Some nits on the test.

PR #161, opened 2023-08-09
feat: use OSTk Data for magnetic files

kyle-cochran:
[Note: this MR is copied almost _vebatim_ from the gravity MR, since the managers are identical]

Modifies the magnetic data manager to be symmetric with other managers. Mainly:

- uses a Mode enum with values Manual/Automatic instead of using an enabled_ flag. They are functionally equivalent
- adds specialized functions to manage the lock file similar to what we have in other managers.
- modifies the actual data fetching function to use the OSTk IO Client::Fetch + add some file integrity checks there.

Much of the functionality that is introduced here can be consolidated into a base class in the near future.
& having this one be a special snowflake is a bit annoying and hard to wrap your head around :).


include/OpenSpaceToolkit/Physics/Environment/Magnetic/Earth/Manager.hpp
```diff
@@ -50,121 +56,122 @@ using EarthMagneticModel = ostk::physics::environment::magnetic::Earth;
...
 class Manager
 {
    public:
-    /// @brief              Copy constructor (deleted)
+    enum class Mode
+    {
 
-    Manager(const Manager& aManager) = delete;
+        Manual,    ///< Manually load and unload gravity files
+        Automatic  ///< Automatically fetch gravity files from remote repositories
```
vishwa2710, line 67:
```suggestion
        Manual,    ///< Manually load and unload magnetic files
        Automatic  ///< Automatically fetch magnetic files from remote repositories
```

vishwa2710 (approved):
Just a nit

PR #162, opened 2023-08-15
refactor: data management refinements from cross-testing in astro

kyle-cochran:
Found 1 bug in space weather fetching and an improperly caught error in frame transformations when doing stuff >1 year in the future.
+ synced the default folders for data files to be symmetric with the structure in OSTk Data


vishwa2710 (approved):
Niceee good catch.

PR #165, opened 2023-08-25
feat: remove hard-coded frames/earth parameters from atmosphere model

kyle-cochran:
We were hard-coding the frame in which atmospheric calculations were being done as well as the earth radius and flattening. These were all used when converting from Position to LLA and back.

This MR exposes them as input to the Earth Atmospheric model constructor, as well as exposing the Sun celestial body input so that the NRLMSISE model can use that in the computation if provided.

Also removes position input from the underlying `Exponential` and `NRLMSISE` models since they only operate on LLA anyway. 

Also adds a loader for the legacy space weather format, mostly for testing purposes now but can be used if we ever want to propagate more than 5 years in the past.

Also adds many more tests :)


include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.hpp
```diff
@@ -47,7 +57,13 @@ class Earth : public Model
...
     /// @param              [in] (optional) aDataDirectory An atmospheric model data directory
 
-    Earth(const Earth::Type& aType, const Directory& aDataDirectory = Directory::Undefined());
+    Earth(
+        const Earth::Type& aType,
+        Shared<const Frame> anEarthFrameSPtr = nullptr,
+        Length anEarthRadius = EarthGravitationalModel::WGS84.equatorialRadius_,
+        Real anEarthFlattening = EarthGravitationalModel::WGS84.flattening_,
+        Shared<Celestial> aSunCelestialSPtr = nullptr
+    );
```
vishwa2710, line 66:
```suggestion
    Earth(
        const Earth::Type& aType,
        const Shared<const Frame>& anEarthFrameSPtr = nullptr,
        const Length& anEarthRadius = EarthGravitationalModel::WGS84.equatorialRadius_,
        const Real& anEarthFlattening = EarthGravitationalModel::WGS84.flattening_,
        const Shared<Celestial>& aSunCelestialSPtr = nullptr
    );
```

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.hpp
```diff
@@ -94,10 +110,9 @@ class Earth : public Model
 
     /// @brief              Get the atmospheric density value at a given position and instant
     ///
-    /// @param              [in] aLLA A position, expressed as latitude, longitude, altitude [deg, deg, m]
-    /// @param              [in] anInstant An instant
+    /// @param              [in] aLLA A LLA
+    /// @param              [in] anInstant An Instant
     /// @return             Atmospheric density value [kg.m^-3]
-
```
vishwa2710, line 100: We should have a space between the docstrings and the function definition.

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.hpp
```diff
@@ -107,7 +122,13 @@ class Earth : public Model
 
     Unique<Impl> implUPtr_;
 
-    static Unique<Impl> ImplFromType(const Type& aType, const Directory& aDataDirectory = Directory::Undefined());
+    static Unique<Impl> ImplFromType(
+        const Type& aType,
+        Shared<const Frame> anEarthFrameSPtr,
+        Length anEarthRadius,
+        Real anEarthFlattening,
+        Shared<Celestial> aSunCelestialSPtr
```
vishwa2710, line 130:
```suggestion
        const Shared<const Frame>& anEarthFrameSPtr,
        const Length& anEarthRadius,
        const Real& anEarthFlattening,
        const Shared<Celestial>& aSunCelestialSPtr
```

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Exponential.hpp
```diff
@@ -38,7 +38,7 @@ using ostk::physics::environment::atmospheric::Model;
 
 /// @brief                      Exponential atmospheric model
 
-class Exponential : public Model
```
vishwa2710, line 41: Is this desired? I thought the architecture was to have all of these inherit from an Atmospheric model base object?

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/NRLMSISE00.hpp
```diff
@@ -30,28 +33,37 @@ using ostk::core::types::Integer;
...
 {
    public:
     /// @brief              Constructor
 
-    NRLMSISE00();
+    NRLMSISE00(
+        Shared<const Frame> anEarthFrameSPtr = nullptr,
+        Length anEarthRadius = EarthGravitationalModel::WGS84.equatorialRadius_,
+        Real anEarthFlattening = EarthGravitationalModel::WGS84.flattening_,
+        Shared<Celestial> aSunCelestialSPtr = nullptr
```
vishwa2710, line 59: idem

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.cpp
```diff
@@ -27,7 +27,12 @@ using EarthGravitationalModel = ostk::physics::environment::gravitational::Earth
 class Earth::Impl
 {
    public:
-    Impl(const Earth::Type& aType, const Directory& aDataDirectory = Directory::Undefined());
+    Impl(
+        const Earth::Type& aType,
```
vishwa2710, line 31: idem

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.cpp
```diff
@@ -57,20 +76,32 @@ Earth::Type Earth::Impl::getType() const
 class Earth::ExponentialImpl : public Earth::Impl
 {
    public:
-    ExponentialImpl(const Earth::Type& aType);
+    ExponentialImpl(
+        const Earth::Type& aType,
+        Shared<const Frame> anEarthFrameSPtr = nullptr,
+        const Length anEarthRadius = EarthGravitationalModel::WGS84.equatorialRadius_,
```
vishwa2710, line 82: idem

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.cpp
```diff
@@ -37,14 +42,28 @@ class Earth::Impl
...
    private:
     Earth::Type type_;
 };
 
-Earth::Impl::Impl(const Earth::Type& aType, const Directory& aDataDirectory)
-    : type_(aType)
+Earth::Impl::Impl(
+    const Earth::Type& aType,
+    Shared<const Frame> anEarthFrameSPtr,
+    const Length anEarthRadius,
```
vishwa2710, line 59: idem

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.cpp
```diff
@@ -57,20 +76,32 @@ Earth::Type Earth::Impl::getType() const
...
    private:
     Exponential exponentialModel_;
 };
 
-Earth::ExponentialImpl::ExponentialImpl(const Earth::Type& aType)
-    : Earth::Impl(aType)
+Earth::ExponentialImpl::ExponentialImpl(
+    const Earth::Type& aType,
+    Shared<const Frame> anEarthFrameSPtr,
+    const Length anEarthRadius,
```
vishwa2710, line 101: idem

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.cpp
```diff
@@ -86,24 +117,51 @@ Real Earth::ExponentialImpl::getDensityAt(const LLA& aLLA, const Instant& anInst
...
+
 class Earth::NRLMSISE00Impl : public Earth::Impl
 {
    public:
-    NRLMSISE00Impl(const Earth::Type& aType);
+    NRLMSISE00Impl(
+        const Earth::Type& aType,
+        Shared<const Frame> anEarthFrameSPtr = nullptr,
+        const Length anEarthRadius = EarthGravitationalModel::WGS84.equatorialRadius_,
+        const Real anEarthFlattening = EarthGravitationalModel::WGS84.flattening_,
```
vishwa2710, line 135: idem

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.cpp
```diff
@@ -86,24 +117,51 @@ Real Earth::ExponentialImpl::getDensityAt(const LLA& aLLA, const Instant& anInst
...
    private:
     NRLMSISE00 NRLMSISE00Model_;
 };
 
-Earth::NRLMSISE00Impl::NRLMSISE00Impl(const Earth::Type& aType)
-    : Earth::Impl(aType)
+Earth::NRLMSISE00Impl::NRLMSISE00Impl(
+    const Earth::Type& aType,
+    Shared<const Frame> anEarthFrameSPtr,
+    const Length anEarthRadius,
```
vishwa2710, line 154: idem

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.cpp
```diff
@@ -118,9 +176,23 @@ Real Earth::NRLMSISE00Impl::getDensityAt(const LLA& aLLA, const Instant& anInsta
...
+    return this->NRLMSISE00Model_.getDensityAt(
+        LLA::Cartesian(aPosition.inFrame(earthFrameSPtr_, anInstant).getCoordinates(), earthRadius_, earthFlattening_),
+        anInstant
+    );
+}
+
+Earth::Earth(
+    const Earth::Type& aType,
+    Shared<const Frame> anEarthFrameSPtr,
+    const Length anEarthRadius,
```
vishwa2710, line 190: idem

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.cpp
```diff
@@ -163,37 +235,35 @@ Earth::Type Earth::getType() const
...
 {
     return implUPtr_->getDensityAt(aLLA, anInstant);
 }
 
-Unique<Earth::Impl> Earth::ImplFromType(const Earth::Type& aType, const Directory& aDataDirectory)
+Unique<Earth::Impl> Earth::ImplFromType(
+    const Earth::Type& aType,
+    Shared<const Frame> anEarthFrameSPtr,
+    const Length anEarthRadius,
+    const Real anEarthFlattening,
```
vishwa2710, line 250: idem

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/NRLMSISE00.cpp
```diff
@@ -52,9 +52,21 @@ using EarthCelestialBody = ostk::physics::env::obj::celest::Earth;
 using EarthGravitationalModel = ostk::physics::environment::gravitational::Earth;
 using ostk::physics::environment::atmospheric::earth::Manager;
 
-NRLMSISE00::NRLMSISE00()
-    : Model()
+NRLMSISE00::NRLMSISE00(
+    Shared<const Frame> anEarthFrameSPtr,
+    const Length anEarthRadius,
```
vishwa2710, line 57: idem

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/NRLMSISE00.cpp
```diff
@@ -185,23 +194,19 @@ Unique<NRLMSISE00::nrlmsise_input> NRLMSISE00::computeNRLMSISE00Input(
     Real lst = Real::Undefined();
 
     // Use actual sun position to compute local solar time if provided
-    if (aSunPosition.isDefined())
+    if (sunCelestialSPtr_)
```
vishwa2710, line 197: While this is a bit simpler, it makes it harder to test as we can't provide the sunPosition as an input...

test/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.test.cpp
```diff
@@ -259,3 +244,111 @@ TEST(OpenSpaceToolkit_Physics_Environment_Atmospheric_Earth, GetDensityAt_LLA)
...
+
+            const Instant instant = std::get<2>(testCase);
+            const Real tolerance = std::get<3>(testCase);
+
+            const Real densityITRF = earthAtmosphericModelITRF.getDensityAt(positionITRF, instant);
+            const Real densityTEME = earthAtmosphericModelTEME.getDensityAt(positionITRF, instant);
+
+            EXPECT_TRUE(densityITRF.isNear(densityTEME, tolerance)) << String::Format(
+                "{} ≈ {} Δ {} [T]", densityITRF.toString(), densityTEME.toString(), (densityITRF - densityTEME)
+            );
```
vishwa2710, line 297: I'm a bit confused. So we're comparing the ITRF density to TEME and seeing that it's within 1e-15?

test/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/NRLMSISE00.test.cpp
```diff
@@ -306,12 +311,10 @@ TEST(OpenSpaceToolkit_Physics_Environment_Atmospheric_Earth_NRLMSISE00, GetDensi
...
             Instant instant = Instant::DateTime(datetime, Scale::UTC);
 
-            const Position sunPosition = sunFrameSPtr->getOriginIn(Frame::ITRF(), instant);
-
-            const Real density = nrlmsise.getDensityAt(position, instant, sunPosition);
+            const Real density = nrlmsise.getDensityAt(lla, instant);
 
             // Check percent tolerance here for low altitudes
-            const Real percentTolerance = 0.5;
+            const Real percentTolerance = 0.42;
```
vishwa2710, line 317: Is this the expected tolerance? Seems a bit high no?

vishwa2710 (changes requested):

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/NRLMSISE00.cpp
kyle-cochran: I don't think it's much harder to test. We just have to initialize two different atmosphere models (one with sun and one without), then compare the difference.

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Exponential.hpp
kyle-cochran: Yeah, this is undoing something that was a huge pain when I first implemented these. 
The problem: we want a super-class for atmosphere models (`Model.hpp`) so that we are able to make atmospheric models for planets that are _not Earth_ and handle them abstractly in the Celestial Object. 

When I implemented these at first, I thought to myself "hey, Exponential and NRLMSISE are the actual models, they should _also_ inherit from this class."

In practice it just means they must have a method `getDensity(Position, Instant)`. At this point I think it makes more sense to just consider these as wrappers that cover the underlying math. And since that math operates on LLA, it doesn't make sense for them to inherit from this `Model.hpp`.

The Earth atmospheric model class `Atmospheric/Earth.hpp` still inherits from `Model`

test/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/NRLMSISE00.test.cpp
kyle-cochran: Yeah this is the best tolerance I could get for these tests.

vishwa2710: Hmmmm okay. Maybe worth discussing this in person I'm a little confused.

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/NRLMSISE00.cpp
vishwa2710: I meant that before, we could actually use the sunposition from other sources (GMAT, Orekit, STK etc.). Now we have can't bootstrap that.

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Exponential.hpp
vishwa2710: Ah gotcha I understand. That makes sense.

test/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.test.cpp
kyle-cochran: Yeah, this is a test to ensure:
1. the input Position is being converted to the correct frame before computing the density
2. to a lesser extent, that our TEME->ITRF conversion doesn't significantly affect the output density

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/NRLMSISE00.cpp
kyle-cochran: Ah yep that's fair. It is harder to validate against another provider when interested specifically in how our computed sun position differs from theirs. 
With the data from yesterday (Thursday), I was able to validate that there is no difference in computed density whether we use our sun position or Orekit's. Which make me more confident that we won't need to do that again. (i.e. we're both just using the sun position provided by SPICE). 
So IMO it's worth the simplicity to add the Sun at the constructor level instead of the function level. I suppose if we need to compare the sun position again, we could mock `getPositionIn` function within the Sun class.

test/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/NRLMSISE00.test.cpp
kyle-cochran: Yeah down to chat more about it

vishwa2710: Oh wait it's 0.42%. Not 42%. 😮‍💨

vishwa2710 (approved):

PR #167, opened 2023-08-28
build: add standalone step for format checker

vishwa2710:


kyle-cochran:
Only comment is that project_name_camel_case should be consistent between this, math, and IO

kyle-cochran (approved):

PR #180, opened 2023-11-02
feat: use predicted manifest update time to determine when to refetch

kyle-cochran:
Pipeline passing depends on the associated OSTk Data MR being merged:
https://github.com/open-space-collective/open-space-toolkit-data/pull/6

Uses the "manifest" entry in the manifest to predict when the remote will have updated data and waits until then to poll for a new manifest file, rather than doing it every time.


src/OpenSpaceToolkit/Physics/Data/Manager.cpp
```diff
@@ -158,12 +156,41 @@ void Manager::setup()
...
+    return File::Path(manifestRepository_.getPath() + Path::Parse(dataManifestFileName)).exists();
+}
+
 void Manager::checkManifestAgeAndUpdate()
 {
-    // Check if the local manifest is too old and fetch a new one if needed
-    // TODO make max age overridable
-    if (!manifestUpdateTimestamp_.isDefined() || !manifest_.isDefined() ||
-        (manifestUpdateTimestamp_ + Duration::Hours(OSTK_PHYSICS_DATA_MANAGER_MANIFEST_MAX_AGE_HOURS) < Instant::Now()))
+    if (!manifest_.isDefined() && !manifestFileExists())
```
kyle-cochran, line 171: This function is the main logical change

include/OpenSpaceToolkit/Physics/Data/Manager.hpp
```diff
@@ -148,6 +153,14 @@ class Manager
...
 
     File fetchLatestManifestFile();
+
+    /// @brief                  Check the age of the manifest to determine if it should be updated before querying it
+    /// for other data files.
+    ///
+    ///                         We fetch the manifest file from the remote if:
+    ///                         - It does not exist locally
+    ///                         - We are past the next predicted update for the manifest based on its own "manifest"
+    ///                         entry.
```
vishwa2710, line 163:
```suggestion

    /// @brief                  Check the age of the manifest to determine if it should be updated before querying it
    /// for other data files.
    ///
    ///                         We fetch the manifest file from the remote if:
    ///                         - It does not exist locally
    ///                         - We are past the next predicted update for the manifest based on its own "manifest"
    ///                         entry.

```

src/OpenSpaceToolkit/Physics/Data/Utilities.cpp
```diff
@@ -0,0 +1,62 @@
...
+        throw ostk::core::error::RuntimeError(
+            "Cannot get file modification time. File [{}] does not exist.", aFile.toString()
+        );
+    }
+
+    std::filesystem::path filePath = std::filesystem::path(std::string(aFile.getPath().getAbsolutePath().toString()));
+
+    const std::filesystem::file_time_type lastWriteTime = std::filesystem::last_write_time(filePath);
+
+    // TBI: this is an approximation. In c++ 20, there is a better way
```
vishwa2710, line 35: You could update physics to use C++ 20. I don't think there's any reason to not upgrade. We will definitely upgrade to 2023 as soon as it's released to use `std::fmt`

vishwa2710 (approved):
LGTM. Just 1 small comment on using C++20

src/OpenSpaceToolkit/Physics/Data/Utilities.cpp
kyle-cochran: Tried it. Looks like it's not a trivial upgrade from 17->20. Some stuff from Core breaks.

vishwa2710: Wow that's surprising. Well, that's a 2024 us problem :P

src/OpenSpaceToolkit/Physics/Data/Manifest.cpp
```diff
@@ -37,13 +39,26 @@ bool Manifest::isDefined() const
     return !dictionary_.isEmpty();
 }
 
+Instant Manifest::getLastUpdateTimestamp() const
```
apaletta3, line 42: why not have the getter function name match the mem var that it is getting from?

apaletta3 (approved):
Couple small comments! Great initiative and implementation.

src/OpenSpaceToolkit/Physics/Data/Manifest.cpp
kyle-cochran: Great point. I've changed this in the MR that will update this one!

src/OpenSpaceToolkit/Physics/Data/Manager.cpp
```diff
@@ -158,12 +156,41 @@ void Manager::setup()
...
+
+        return;
+    }
+
+    if (!manifest_.isDefined() && manifestFileExists())
+    {
+        // A manifest file exists but we haven't loaded it yet. Load it.
+        this->loadManifest(Manifest::Load(File::Path(manifestRepository_.getPath() + Path::Parse(dataManifestFileName)))
+        );
+    }
```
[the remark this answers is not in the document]
kyle-cochran: Good looks. 👍 an if/elseif pattern is more appropriate here.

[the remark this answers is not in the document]
apaletta3: wait what? which one is more appropriate? You currently don't have an if/elseif pattern.

Issue #179, opened 2023-10-31
[feat] add shortcut environment setting to configure all data at once


kyle-cochran:
**Is your feature request related to a problem? Please describe.**
To configure where OSTk Data is stored, you currently have to set the path for each individual data set separately:
```Dockerfile
ENV OSTK_LOCAL_CACHE="${OSTK_LOCAL_CACHE}"

ENV OSTK_PHYSICS_DATA_MANIFEST_LOCAL_REPOSITORY="${OSTK_LOCAL_CACHE}"
ENV OSTK_PHYSICS_COORDINATE_FRAME_PROVIDERS_IERS_MANAGER_LOCAL_REPOSITORY="${OSTK_PHYSICS_DATA_MANIFEST_LOCAL_REPOSITORY}/coordinate/frame/providers/iers"
ENV OSTK_PHYSICS_ENVIRONMENT_EPHEMERIDES_SPICE_MANAGER_LOCAL_REPOSITORY="${OSTK_PHYSICS_DATA_MANIFEST_LOCAL_REPOSITORY}/environment/ephemerides/spice"
ENV OSTK_PHYSICS_ENVIRONMENT_GRAVITATIONAL_EARTH_MANAGER_LOCAL_REPOSITORY="${OSTK_PHYSICS_DATA_MANIFEST_LOCAL_REPOSITORY}/environment/gravitational/earth"
ENV OSTK_PHYSICS_ENVIRONMENT_MAGNETIC_EARTH_MANAGER_LOCAL_REPOSITORY="${OSTK_PHYSICS_DATA_MANIFEST_LOCAL_REPOSITORY}/environment/magnetic/earth"
ENV OSTK_PHYSICS_ENVIRONMENT_ATMOSPHERIC_EARTH_MANAGER_LOCAL_REPOSITORY="${OSTK_PHYSICS_DATA_MANIFEST_LOCAL_REPOSITORY}/environment/atmospheric/earth"
```

This can very cumbersome, and if you miss any, the ones you miss are set to a default location. I don't think I've ever seen a situation where they needed to be different.

**Describe the solution you'd like**
One environment variable to rule them all. `OSTK_PHYSICS_DATA_LOCAL_REPOSITORY`

This will set all of the OSTk data locations to the given path. However, if any of the above _specific_ env vars are set, they should override this one.

kyle-cochran:
Implemented as `OSTK_PHYSICS_DATA_LOCAL_REPOSITORY`

PR #181, opened 2023-11-02
feat: add OSTK_PHYSICS_DATA_LOCAL_REPOSITORY environment setting to set all at once

kyle-cochran:
When installing OSTk into a new environment, if the default data files path isn't to your liking, you have to set 7 environment variable to make sure all of the data are put into the correct place. This is a bit annoying and causes problem when new data are added to physics, because users do not know to add the new settings.

This adds a single configuration env var called `OSTK_PHYSICS_DATA_LOCAL_REPOSITORY` that let's you set the location for all data at once. The original environment variable are still available and override this one.

Addresses: https://github.com/open-space-collective/open-space-toolkit-physics/issues/179


apaletta3 (approved):
Epic kyle! Great idea!

PR #183, opened 2023-11-07
refactor!: remove unused in-memory caches from IERS data manager

kyle-cochran:
The IERS data manager has a cache so that multiple sets of each type of IERS data (bulletin A/Finals 2000A) may be loaded into memory at the same time. It's currently unused (except for some extremely unlikely edge cases) and not needed (best I can tell, even before the data refactor it wasn't actually being utilized).

It make the logic for data management really mind-bending and hard to maintain or alter. It was also causing a bug in [this MR](https://github.com/open-space-collective/open-space-toolkit-physics/pull/180) that made us fetch the manifest file almost infinitely 😬.

Since we don't use it, I'd like to take the opportunity to remove this idiosyncrasy and make things a little simpler.

This also uncovered a bug in how we parse the Bulletin A data, which was causing us to parse some observational data as predictive.

This also uncovered a second bug where the Reset method just didn't really work. (the test for it was skipped)

**breaking**: this breaks a couple of methods exposed in the IERS data Manager that are related to the caching. I'd be extremely surprised if they're actually use anywhere outside of Physics.


src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Manager.cpp
```diff
@@ -207,17 +195,11 @@ Real Manager::getUt1MinusUtcAt(const Instant& anInstant) const
...
-                "Cannot obtain UT1 - UTC from Bulletin A at [{}].", anInstant.toString()
-            );
-        }
     }
 
-    const Finals2000A* finals2000aPtr = this->accessFinals2000AAt(anInstant);
+    const Finals2000A* finals2000aPtr = this->accessFinals2000A();
 
-    if (finals2000aPtr != nullptr)
+    if (finals2000aPtr)
```
vishwa2710, line 202:
```suggestion
    if (finals2000aPtr != nullptr)
```
Nit: I believe this is more explicit and thus preferred.

include/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Manager.hpp
```diff
@@ -95,52 +95,38 @@ class Manager
...
+    BulletinA getBulletinA();
 
-    /// @brief              Get Finals 2000A at instant
+    /// @brief              Get Finals 2000A
     ///
-    /// @param              [in] anInstant An instant
     /// @return             Finals 2000A
 
-    Finals2000A getFinals2000AAt(const Instant& anInstant) const;
+    Finals2000A getFinals2000A();
```
vishwa2710, line 108:
```suggestion
    Finals2000A getFinals2000A() const;
```
I believe it's worth marking these as const, as the relevant member variables that changed are marked as mutable.

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Manager.cpp
```diff
@@ -314,14 +296,9 @@ void Manager::reset()
...
-    aBulletinIndex_ = 0;
-    finals2000aIndex_ = 0;
-
-    bulletinAUpdateTimestamp_ = Instant::Undefined();
-    finals2000AUpdateTimestamp_ = Instant::Undefined();
-
-    aBulletins_.clear();
-    finals2000aArray_.clear();
+    bulletinA_ = BulletinA::Undefined();
+    // std::cout << bulletinA_ << std::endl;
```
vishwa2710, line 300:
```suggestion
```

vishwa2710 (approved):
Just a few nits.

include/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Manager.hpp
```diff
@@ -227,26 +213,20 @@ class Manager
...
 
-    mutable Index aBulletinIndex_;
-    mutable Index finals2000aIndex_;
-
-    mutable Instant bulletinAUpdateTimestamp_;
-    mutable Instant finals2000AUpdateTimestamp_;
-
     Manager(const Manager::Mode& aMode = Manager::DefaultMode());
 
     bool isLocalRepositoryLocked() const;
```
apaletta3, line 223: Maybe all or none of the private mem funcs here should follow the ending with underscore_ naming pattern

include/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Manager.hpp
apaletta3: But they're not marked as mutable anymore, it seems Kyle removed those variables? And having the "intention" shown at the interface level is better than lying at the interface and then allowing mutation of the private vars right?

include/OpenSpaceToolkit/Physics/Data/Manager.hpp
```diff
@@ -152,7 +152,8 @@ class Manager
 
     void setup();
 
-    File fetchLatestManifestFile();
+    // virtual so we can mock in testing
```
apaletta3, line 155: Make const if possible

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Finals2000A.cpp
```diff
@@ -450,7 +442,8 @@ Finals2000A Finals2000A::Load(const fs::File& aFile)
 }
 
 Finals2000A::Finals2000A()
```
apaletta3, line 444: can you move the constructor to the top of the file and make sure the method order follows our pattern?

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Finals2000A.cpp
```diff
@@ -450,7 +442,8 @@ Finals2000A Finals2000A::Load(const fs::File& aFile)
...
-    : span_(Interval::Undefined()),
+    : lastModifiedTimestamp_(Instant::Undefined()),
+      span_(Interval::Undefined()),
       data_(Map<Real, Finals2000A::Data>())
 {
 }
```
apaletta3, line 449: Seems that we return raw pointers here instead of references? Could be worth fixing just to be safer

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Manager.cpp
```diff
@@ -82,55 +82,35 @@ Directory Manager::getFinals2000ADirectory() const
...
-
     std::lock_guard<std::mutex> lock {mutex_};
 
-    const Finals2000A* finals2000aPtr = this->accessFinals2000AAt(anInstant);
+    const Finals2000A* finals2000aPtr = this->accessFinals2000A();
 
     if (finals2000aPtr != nullptr)
     {
         return *finals2000aPtr;
     }
```
apaletta3, line 108: Should these be made Unique pointers (or shared) instead of raw pointers? These can still be compared to `nullptr` if I remember correctly

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Manager.cpp
```diff
@@ -414,159 +387,112 @@ bool Manager::isLocalRepositoryLocked() const
...
     {
-        const BulletinA& bulletinA = aBulletins_.at(aBulletinIndex_);
-
-        if (bulletinA.accessObservationInterval().contains(anInstant) ||
-            bulletinA.accessPredictionInterval().contains(anInstant
-            ))  // [TBI] Check that next observation bulletin available first
-        {
-            return &bulletinA;
-        }
+        return &bulletinA_;
```
apaletta3, line 395: ditto here about returning raw pointers, why not references?

apaletta3 (changes requested):
Left a few comments about naming and use of raw pointers and pointers in return values, but otherwise looks good.

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Manager.cpp
kyle-cochran: I'm fine with either. Though we had decided on the other though 🤔

include/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Manager.hpp
kyle-cochran: This is a really interesting thing to bring up. Thanks for calling it out @vishwa2710. I looked into why and when you should use `mutable` with `const` and it has a lot to do with concurrency in C++.

Really good article about it written by the chair of the C++ standards committee:
https://herbsutter.com/2013/05/24/gotw-6a-const-correctness-part-1-3/

It mostly boils down to:

> Using const consistently is simply necessary for correctly-synchronized code.

> If your class contains a member that could change even for const objects and operations, make that member mutable and protect it with a mutex or make it atomic.

Having `const` member functions is important for parallel access to the `Manager` object. It allows multiple threads to know that they can access the object concurrently. Writes to an object by `const` member functions is okay but only if the member is mutable and protected by a mutex (called the M&M rule).

So at the end of the day, I do think we want these member functions to be marked `const`, but we **do not** want to `const_cast` everywhere like we were doing in some places. This should be handled in the proper way by marking things as mutable.

TIL 🌈

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Finals2000A.cpp
kyle-cochran: I agree about this. Want to avoid a [chestertons fence](https://fs.blog/chestertons-fence/) and not change too many things at once. Cool if this is in a follow-up?

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Finals2000A.cpp
kyle-cochran: I might just not be savvy here, but do we have an established pattern? 

For this function, the constructor is private, and all object construction is done through the `Load` factory function. This is a pattern that exists in the Manager classes, too. In these cases I think it's appropriate to have the constructor definition on the bottom since it matches the header file.

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Manager.cpp
kyle-cochran: I was super inconsistent in where I changed this anyway. Reverted!

include/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Manager.hpp
apaletta3: Yup okay agreed

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Finals2000A.cpp
apaletta3: Ah yes sorry I read the file too quick... I thought it was a public constructor so I was wondering why it was at the bottom, but I see the factory patter now. Cool!

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Finals2000A.cpp
apaletta3: yup!

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Manager.cpp
apaletta3: note to self, can be fixed in future ^

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Providers/IERS/Manager.cpp
apaletta3: note to self, can be fixed in future ^

apaletta3 (approved):
Noice

PR #184, opened 2023-11-08
feat: add environment variable OSTK_PHYSICS_DATA_REFRESH_RATE_H to throttle non-critical data fetching

kyle-cochran:
Adds an environment variable to prevent OSTk from fetching a new manifest file to check for updates unless it is some number of hours old. 

Importantly, critical IO will still happen. If spaceweather data is needed and we don't have the file, it should still be fetched. However, if the spaceweather file is 2 days old and the data refresh rate is set to 3 days, we won't fetch it.

### DIFF
--- a/include/OpenSpaceToolkit/Physics/Data/Manager.hpp
+++ b/include/OpenSpaceToolkit/Physics/Data/Manager.hpp
@@ -172,11 +172,10 @@ class Manager
 
     static Duration DefaultManifestRepositoryLockTimeout();
 
-    /// @brief                  Determines how often OSTk should do non-critical IO to freshen its data.
+    /// @brief                  Determine how often OSTk should do non-critical IO to freshen its data.
     /// We limit this by limiting how frequently the manifest file is updated.
     ///
-    ///
-    /// @return                 Frequency in hours. Default is 24. Overriden by the OSTK_PHYSICS_DATA_REFRESH_RATE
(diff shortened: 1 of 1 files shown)


include/OpenSpaceToolkit/Physics/Data/Manager.hpp
```diff
@@ -171,6 +171,15 @@ class Manager
     void unlockManifestRepository();
 
     static Duration DefaultManifestRepositoryLockTimeout();
+
+    /// @brief                  Determines how often OSTk should do non-critical IO to freshen its data.
+    /// We limit this by limiting how frequently the manifest file is updated.
+    ///
+    ///
+    /// @return                 Frequency in hours. Default is 24. Overriden by the OSTK_PHYSICS_DATA_REFRESH_RATE
+    /// environment variable.
```
vishwa2710, line 180:
```suggestion
    /// @brief                  Determines how often OSTk should do non-critical IO to freshen its data.
    /// We limit this by limiting how frequently the manifest file is updated.
    ///
    /// @return                 Frequency as a duration. Default is 24 hours. Overriden by the OSTK_PHYSICS_DATA_REFRESH_RATE
    /// environment variable.
```

vishwa2710 (approved):

Derollez (approved):

PR #185, opened 2023-11-08
feat: add fixed solar flux and geomagnetic index input type to Earth atmospheric density models

Derollez:
- [x] A couple of improvements on default values. 
- [x] Default args in python.
- [x] Reduce breaking changes if possible (default args). 
- [ ] Tutorial to show usage.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth.cpp
+++ b/bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth.cpp
@@ -86,7 +86,7 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth(pybind11::mo
 
             .value("Undefined", Earth::InputDataSourceType::Undefined)
             .value("ConstantFluxAndGeoMag", Earth::InputDataSourceType::ConstantFluxAndGeoMag)
-            .value("SpaceWeatherFile", Earth::InputDataSourceType::SpaceWeatherFile);
+            .value("CSSISpaceWeatherFile", Earth::InputDataSourceType::CSSISpaceWeatherFile);
     }
 
     // Create "earth" python submodule
--- a/bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth.cpp
+++ b/bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth.cpp
@@ -35,7 +35,7 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth(pybind11::mo
             .def(
                 init<
                     const Earth::Type&,
-                    const Earth::InputDataSourceType&,
+                    const Earth::InputDataType&,
                     const Real&,
                     const Real&,
                     const Real&,
@@ -44,7 +44,7 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth(pybind11::mo
                     const Real&,
                     const Shared<Celestial>&>(),
                 arg("type"),
-                arg("input_data_source_type"),
+                arg("input_data_type"),
                 arg("f107_constant_value") = 150.0,
                 arg("f107_average_constant_value") = 150.0,
                 arg("kp_constant_value") = 3.0,
@@ -56,7 +56,7 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth(pybind11::mo
 
             .def("get_type", &Earth::getType)
 
-            .def("get_input_data_source_type", &Earth::getInputDataSourceType)
+            .def("get_input_data_type", &Earth::getInputDataType)
 
             .def("is_defined", &Earth::isDefined)
 
@@ -82,11 +82,11 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth(pybind11::mo
             .value("Exponential", Earth::Type::Exponential)
             .value("NRLMSISE00", Earth::Type::NRLMSISE00);
 
-        enum_<Earth::InputDataSourceType>(earth_class, "InputDataSourceType")
+        enum_<Earth::InputDataType>(earth_class, "InputDataType")
 
-            .value("Undefined", Earth::InputDataSourceType::Undefined)
-            .value("ConstantFluxAndGeoMag", Earth::InputDataSourceType::ConstantFluxAndGeoMag)
-            .value("CSSISpaceWeatherFile", Earth::InputDataSourceType::CSSISpaceWeatherFile);
+            .value("Undefined", Earth::InputDataType::Undefined)
+            .value("ConstantFluxAndGeoMag", Earth::InputDataType::ConstantFluxAndGeoMag)
+            .value("CSSISpaceWeatherFile", Earth::InputDataType::CSSISpaceWeatherFile);
     }
 
     // Create "earth" python submodule
--- a/bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth.cpp
+++ b/bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth.cpp
@@ -35,11 +35,19 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth(pybind11::mo
             .def(
                 init<
                     const Earth::Type&,
+                    const Earth::InputDataSourceType&,
+                    const Real&,
+                    const Real&,
+                    const Real&,
                     const Shared<const Frame>&,
                     const Length&,
                     const Real&,
                     const Shared<Celestial>&>(),
                 arg("type"),
+                arg("input_data_source_type"),
+                arg("f107_constant_value") = 150.0,
+                arg("f107_average_constant_value") = 150.0,
+                arg("kp_constant_value") = 3.0,
                 arg("earth_frame") = Frame::ITRF(),
                 arg("earth_radius") = EarthGravityModel::WGS84.equatorialRadius_,
                 arg("earth_flattening") = EarthGravityModel::WGS84.flattening_,
@@ -47,6 +55,8 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth(pybind11::mo
             )
 
             .def("get_type", &Earth::getType)
+
+            .def("get_input_data_source_type", &Earth::getInputDataSourceType)
 
             .def("is_defined", &Earth::isDefined)
 
@@ -71,6 +81,12 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth(pybind11::mo
             .value("Undefined", Earth::Type::Undefined)
             .value("Exponential", Earth::Type::Exponential)
             .value("NRLMSISE00", Earth::Type::NRLMSISE00);
+
+        enum_<Earth::InputDataSourceType>(earth_class, "InputDataSourceType")
+
+            .value("Undefined", Earth::InputDataSourceType::Undefined)
+            .value("ConstantFluxAndGeoMag", Earth::InputDataSourceType::ConstantFluxAndGeoMag)
+            .value("SpaceWeatherFile", Earth::InputDataSourceType::SpaceWeatherFile);
     }
 
     // Create "earth" python submodule
--- a/bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth.cpp
+++ b/bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth.cpp
@@ -30,6 +30,18 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth(pybind11::mo
     {
         class_<Earth, Shared<Earth>> earth_class(aModule, "Earth");
 
+        enum_<Earth::Type>(earth_class, "Type")
+
+            .value("Undefined", Earth::Type::Undefined)
+            .value("Exponential", Earth::Type::Exponential)
+            .value("NRLMSISE00", Earth::Type::NRLMSISE00);
+
+        enum_<Earth::InputDataType>(earth_class, "InputDataType")
+
+            .value("Undefined", Earth::InputDataType::Undefined)
+            .value("ConstantFluxAndGeoMag", Earth::InputDataType::ConstantFluxAndGeoMag)
+            .value("CSSISpaceWeatherFile", Earth::InputDataType::CSSISpaceWeatherFile);
+
         earth_class
 
             .def(
@@ -44,10 +56,10 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth(pybind11::mo
                     const Real&,
                     const Shared<Celestial>&>(),
                 arg("type"),
-                arg("input_data_type"),
-                arg("f107_constant_value") = 150.0,
-                arg("f107_average_constant_value") = 150.0,
-                arg("kp_constant_value") = 3.0,
+                arg("input_data_type") = Earth::InputDataType::Undefined,
+                arg("f107_constant_value") = Earth::defaultF107ConstantValue,
+                arg("f107_average_constant_value") = Earth::defaultF107AConstantValue,
+                arg("kp_constant_value") = Earth::defaultKpConstantValue,
                 arg("earth_frame") = Frame::ITRF(),
                 arg("earth_radius") = EarthGravityModel::WGS84.equatorialRadius_,
                 arg("earth_flattening") = EarthGravityModel::WGS84.flattening_,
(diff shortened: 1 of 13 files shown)


bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth.cpp
```diff
@@ -35,11 +35,19 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth(pybind11::mo
...
+                    const Real&,
                     const Shared<const Frame>&,
                     const Length&,
                     const Real&,
                     const Shared<Celestial>&>(),
                 arg("type"),
+                arg("input_data_source_type"),
+                arg("f107_constant_value") = 150.0,
+                arg("f107_average_constant_value") = 150.0,
+                arg("kp_constant_value") = 3.0,
```
kyle-cochran, line 50: It may be cleaner to use the `Ap` scale rather than the `Kp` scale here. We use `Ap ` in the computations, and I believe `Kp` has a bunch of weird qualifiers that would be awkward to parse in the input:
https://www.ngdc.noaa.gov/stp/geomag/kp_ap.html

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.hpp
```diff
@@ -49,16 +49,34 @@ class Earth : public Model
...
         Exponential,  ///< Exponential atmospheric density model, valid up to 1000 km
-        NRLMSISE00    ///< Navy Research Lab Mass Spectrometer and Incoherent Scatter Radar Exosphere 2000
+        NRLMSISE00,   ///< Navy Research Lab Mass Spectrometer and Incoherent Scatter Radar Exosphere 2000
+    };
+
+    enum class InputDataSourceType
+    {
+        Undefined,              ///< Undefined
+        ConstantFluxAndGeoMag,  ///< Use constant values for F10.7, F10.7a and Kp NRLMSISE00 input parameters
+        SpaceWeatherFile,  ///< Use historical and predicted values for F10.7, F10.7a and Kp NRLMSISE00 input parameters
```
kyle-cochran, line 59:
```suggestion
        CSSISpaceWeatherFile,  ///< Use historical and predicted values for F10.7, F10.7a and Kp NRLMSISE00 input parameters
```
nit: can we name this with `CSSI`? If we have a `Schatten` in the future, that will also presumably be a space-weather file. Also, CSSI is more searchable for someone trying to understand what's happening

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.hpp
```diff
@@ -113,8 +137,13 @@ class Earth : public Model
     /// @param              [in] aLLA A LLA
     /// @param              [in] anInstant An Instant
     /// @return             Atmospheric density value [kg.m^-3]
+
     Real getDensityAt(const LLA& aLLA, const Instant& anInstant) const;
 
+    static constexpr double defaultF107ConstantValue = 150.0;   // 10⁻²² W⋅m⁻²⋅Hz⁻¹
```
kyle-cochran, line 143: nit: defaults in Physics usually use header defines rather than `constexpr`
https://github.com/open-space-collective/open-space-toolkit-physics/blob/main/include/OpenSpaceToolkit/Physics/Time/Instant.hpp#L29

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.cpp
```diff
@@ -253,13 +308,59 @@ Unique<Earth::Impl> Earth::ImplFromType(
...
+        }
+        else if (anInputDataSourceType == Earth::InputDataSourceType::SpaceWeatherFile)
+        {
+            return std::make_unique<NRLMSISE00Impl>(
+                aType,
+                anInputDataSourceType,
+                NRLMSISE00::InputDataSourceType::SpaceWeatherFile,
+                aF107ConstantValue,
+                aF107AConstantValue,
+                aKpConstantValue,
```
kyle-cochran, line 357:
```suggestion
```

mega nit: I think this is a little clearer that they are unused

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth.cpp
kyle-cochran: Re: slack thread on this. GMAT uses Kp and it seems freeflyer allows either:
https://ai-solutions.com/_help_Files/user_created_atmospheric_drag_files.htm

Let's stick with your implementation!

kyle-cochran: I think the first link I posted has a different definition of Kp than some other sources, where it just goes from 0-9

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth/NRLMSISE00.cpp
```diff
@@ -15,21 +16,44 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth_NRLMSISE00(p
     using ostk::physics::coord::Frame;
     using ostk::physics::coord::spherical::LLA;
     using ostk::physics::env::obj::Celestial;
+    using EarthGravityModel = ostk::physics::environment::gravitational::Earth;
+    // using EarthAtmosphericModel = ostk::physics::environment::atmospheric::Earth;
```
kyle-cochran, line 20:
```suggestion
```

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.hpp
```diff
@@ -49,16 +49,34 @@ class Earth : public Model
     {
         Undefined,    ///< Undefined
         Exponential,  ///< Exponential atmospheric density model, valid up to 1000 km
-        NRLMSISE00    ///< Navy Research Lab Mass Spectrometer and Incoherent Scatter Radar Exosphere 2000
+        NRLMSISE00,   ///< Navy Research Lab Mass Spectrometer and Incoherent Scatter Radar Exosphere 2000
+    };
+
+    enum class InputDataSourceType
```
kyle-cochran, line 55:
```suggestion
    enum class InputDataType
```
nit: I think this could be shortened to either `InputDataType` or `InputDataSource` and still be clear IMO.

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.cpp
```diff
@@ -253,13 +308,59 @@ Unique<Earth::Impl> Earth::ImplFromType(
...
+        );
     }
     else if (aType == Earth::Type::NRLMSISE00)
     {
-        return std::make_unique<NRLMSISE00Impl>(
-            aType, anEarthFrameSPtr, anEarthRadius, anEarthFlattening, aSunCelestialSPtr
-        );
+        if (anInputDataSourceType == Earth::InputDataSourceType::Undefined)
+        {
+            throw ostk::core::error::runtime::Undefined("Input data source type");
```
kyle-cochran, line 319:
```suggestion
            Earth::InputDataSourceType defaultNRLMSISEDataSource = Earth::InputDataSourceType::SpaceWeatherFile;
            
            return std::make_unique<NRLMSISE00Impl>(
                aType,
                anInputDataSourceType,
                NRLMSISE00::InputDataSourceType::SpaceWeatherFile,
                anEarthFrameSPtr,
                anEarthRadius,
                anEarthFlattening,
                aSunCelestialSPtr
            );
```

Maybe we could do something like this instead of throwing an error here. Using the space weather file seems like a reasonable default to me and matches current behavior.

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/NRLMSISE00.cpp
```diff
@@ -77,100 +85,148 @@ bool NRLMSISE00::isDefined() const
...
-    // Fetch AP indices for all days. Stack them into one continuous array.
-    Array<Integer> apMultiDayArray = Array<Integer>::Empty();
-    apMultiDayArray.reserve(fetchDays.getSize());
+    Unique<NRLMSISE00::ap_array> outputStruct = std::make_unique<NRLMSISE00::ap_array>();
 
-    for (const Instant& fetchDay : fetchDays)
+    switch (this->inputDataSourceType_)
     {
-        apMultiDayArray.add(spaceWeatherManager.getAp3HourSolarIndicesAt(fetchDay));
+        case InputDataSourceType::ConstantFluxAndGeoMag:
```
kyle-cochran, line 92: I think this logic can be simplified a bit. The AP struct is actually allowed to be a null pointer if we're not using it. (unless a particular switch is set to -1, which we don't do)
https://github.com/magnific0/nrlmsise-00/blob/master/nrlmsise-00.h#L50

So you can probably skip the entire `computeApArray` function for the case of constant input.

kyle-cochran: I can't comment on it directly since it's not changed, but we'll need to set this line in the input struct to point to the value of our constant Ap:
> input->ap_a = apValues.get();

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth.cpp
Derollez: Hum, interesting. I'll double check those 🙏

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.hpp
Derollez: Yeah good point. I wanted to make that change when we would have added a new type like Schatten later but you're right: that will be yet another breaking change, better now 💯

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.cpp
Derollez: I like that suggestion! thanks!

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.hpp
Derollez: Done

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.hpp
Derollez: I checked real quick with chatgpt, apparently
"
With this definition, every occurrence of DEFAULT_F107_CONSTANT_VALUE in your code will be replaced with 150.0 during the preprocessing stage. This is a simple textual substitution and doesn't have the type safety or scoping of true C++ constants.
For a more C++-style approach to defining constants with type safety, it's recommended to use const variables or, in C++11 and later, constexpr variables
" 🤷 Lmk but I'll keep it as is for now and we can update batch on the repo later

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.cpp
Derollez: Yeah, we would need `Earth::NRLMSISE00Impl::NRLMSISE00Impl` to accept default args for this (and probably change the order of args though 🤔)

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth.cpp
Derollez: Okay, following your recommendation below, closing this one

include/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.hpp
Derollez: Yeah 💯

Derollez: applied

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.cpp
Derollez: Applied. Default for NRLMSISE and Undefined will be CSSISpaceWeatherFile

bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth/NRLMSISE00.cpp
Derollez: Re-added because I am re-using it for default values

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth.cpp
Derollez: Even better, with the default values in python, no breaking changes and combined with your suggestion here:
- Exponential > Undefined. all good
- Exponential > Other. all good
- NRLMSISE > Undefined > meaning CSSI. All good. We can keep the same behavior we had before
- NRLMSISE > CSSI. All good
- NRLMSISE > Constant. Then all good

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/NRLMSISE00.cpp
Derollez: I tried the following
```
Unique<NRLMSISE00::nrlmsise_input> NRLMSISE00::computeNRLMSISE00Input(
    const Unique<NRLMSISE00::ap_array>& apValues, const LLA& aLLA, const Instant& anInstant
) const
{
    Real f107Previous = Real::Undefined();
    Real f107Average = Real::Undefined();

    switch (this->inputDataType_)
    {
        case InputDataType::ConstantFluxAndGeoMag:
        {
            // Use constant values for F10.7, F10.7a and Kp NRLMSISE00 input parameters

            f107Previous = this->f107ConstantValue_;
            f107Average = this->f107AConstantValue_;

            break;
        }

        case InputDataType::CSSISpaceWeatherFile:
        {
            // Use historical and predicted values for F10.7, F10.7a and Kp NRLMSISE00 input parameters

            // Input reference is in the NRLMSISE header file
            // https://github.com/magnific0/nrlmsise-00/blob/master/nrlmsise-00.h

            const Manager& spaceWeatherManager = Manager::Get();

            // Solar flux values
            f107Previous = spaceWeatherManager.getF107SolarFluxAt(anInstant - Duration::Days(1));
            f107Average = spaceWeatherManager.getF107SolarFlux81DayAvgAt(anInstant);

            break;
        }
    }

    const DateTime currentDateTime = anInstant.getDateTime(Scale::UTC);

    // current year/doy/sec
    const Integer year = currentDateTime.getDate().getYear();

    const Instant startOfYear = Instant::DateTime(DateTime(Date(year, 1, 1), Time::Midnight()), Scale::UTC);
    const Integer dayOfYear = (anInstant - startOfYear).getDays() + 1;

    const Time timeOfDay = currentDateTime.getTime();
    const Integer secondsInDay = timeOfDay.getHour() * 3600 + timeOfDay.getMinute() * 60 + timeOfDay.getSecond();

    Real lst = Real::Undefined();

    // Use actual sun position to compute local solar time if provided
    if (sunCelestialSPtr_)
    {
        const Position position = {
            aLLA.toCartesian(earthRadius_, earthFlattening_), Position::Unit::Meter, earthFrameSPtr_
        };

        const Position sunPositionInFrame = sunCelestialSPtr_->getPositionIn(earthFrameSPtr_, anInstant);

        lst = (Real::Pi() + std::atan2(
                                sunPositionInFrame.accessCoordinates()[0] * position.accessCoordinates()[1] -
                                    sunPositionInFrame.accessCoordinates()[1] * position.accessCoordinates()[0],
                                sunPositionInFrame.accessCoordinates()[0] * position.accessCoordinates()[0] +
                                    sunPositionInFrame.accessCoordinates()[1] * position.accessCoordinates()[1]
                            )) *
              12.0 / Real::Pi();
    }
    else
    {
        // This is the preferred method per the NRLMSISE documentation
        // https://github.com/magnific0/nrlmsise-00/blob/master/nrlmsise-00.h#L103
        lst = Real::Integer(secondsInDay) / 3600.0 + aLLA.getLongitude().inDegrees() / 15.0;
    }

    Unique<NRLMSISE00::nrlmsise_input> input = std::make_unique<NRLMSISE00::nrlmsise_input>();

    input->doy = dayOfYear;
    input->year = year;
    input->sec = secondsInDay;
    input->alt = aLLA.getAltitude().inKilometers();
    input->g_lat = aLLA.getLatitude().inDegrees();
    input->g_long = aLLA.getLongitude().inDegrees();
    input->lst = lst;
    input->f107A = f107Average;
    input->f107 = f107Previous;

    if (apValues != nullptr)
    {
        input->ap = apValues->a[0];
        input->ap_a = apValues.get();
    }
    else
    {
        const Real ap = this->convertKpToAp(this->kpConstantValue_);

        input->ap = ap;
        input->ap_a = nullptr;
    }

    return input;
}

Real NRLMSISE00::convertKpToAp(const Real& aKp)
{
    if (!aKp.isDefined())
    {
        throw ostk::core::error::runtime::Undefined("Kp");
    }

    return exp((aKp + 1.6) / 1.75);  // Exponential approximation, Vallado, 2nd edition (eq 8-31).
}

NRLMSISE00::InputDataType NRLMSISE00::getInputDataType() const
{
    return this->inputDataType_;
}

Real NRLMSISE00::getF107ConstantValue() const
{
    return this->f107ConstantValue_;
}

Real NRLMSISE00::getF107AConstantValue() const
{
    return this->f107AConstantValue_;
}

Real NRLMSISE00::getKpConstantValue() const
{
    return this->kpConstantValue_;
}

Real NRLMSISE00::getDensityAt(const LLA& aLLA, const Instant& anInstant) const
{
    Unique<NRLMSISE00::ap_array> apValues = nullptr;

    if (this->inputDataType_ == InputDataType::CSSISpaceWeatherFile)
    {
        apValues = this->computeApArray(anInstant);
    }

    const Unique<NRLMSISE00::nrlmsise_input> input = this->computeNRLMSISE00Input(apValues, aLLA, anInstant);

    return NRLMSISE00::GetDensityAt(*input);
}
```

I must be doing something wrong 🤔

Derollez: lmk

kyle-cochran (approved):

PR #191, opened 2023-11-22
refactor!: rename abbreviated namespaces to full pathing for Core/IO/Math

vishwa2710:


lucas-bremond:
This should be a `refactor!` no?

vishwa2710:
Absolutely it should!
I always forget about the ! :'(

PR #201, opened 2024-02-07
chore: suggestions to propagate the ephemerides->ephemeris change

kyle-cochran:
This changes the name of the Data environment variables (which I would be surprised if anyone used) and the _local_ storage place for files.

It does not change the place on the remote where it looks for data. That change does not break the interface, so I'll do it in a separate MR.


test/OpenSpaceToolkit/Physics/Environment/Ephemerides/SPICE/AccessFrame/Scenario_1/Scenario_1 STK ScenarioWB.wsp
```diff
@@ -24,7 +24,7 @@ ptn_Child7=Document-6
 [WorkState_v1_2.Frames.ChildFrames.Document-0]
 ptn_Child1=ViewFrame-0
 [WorkState_v1_2.Frames.ChildFrames.Document-0.ViewFrame-0]
-DocPathName=Z:\Projects\Engineering\open-space-collective\open-space-toolkit-physics\test\OpenSpaceToolkit\Physics\Environment\Ephemerides\SPICE\AccessFrame\AgUi3DGfxView11.AgUiPi3DGfxView.1
```
kyle-cochran, line 27: I don't think this file matters but just in case

vishwa2710 (approved):

PR #204, opened 2024-02-14
feat: use OSTk Data V1 API to align file locations

kyle-cochran:
After changing the namespace and filenames in OSTk, the structure of data in OSTk Physics no longer mirrored OSTk Data. It was not safe to simply update OSTk Data to use the new structure, as it would break any OSTk that was not updated to at least Physics V5. 

To solve this, we introduced versioning into the OSTk Data "API". The versioning is present in the URL that is used to fetch data:

this: https://github.com/open-space-collective/open-space-toolkit-data/blob/main/data/manifest.json
vs. this: https://github.com/open-space-collective/open-space-toolkit-data/blob/v1/data/manifest.json

Which is implemented by just created a new synced branch on the OSTk Data remote, called `v1`.

This MR does 2 things:
1.) change hard-coded remote paths to use the manifest as a reference (should've been done a long time ago)
2.) update OSTk Data URL to the new V1 API, so that the local file structure matches the remote.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,9 @@ The library exhibits the following structure:
 </details>
 
 ## Data Management
+
 OSTk Physics uses input data from various sources to determine the state of the space environment at any given time. The following input data is used:
+
 - IERS Reference Frame Data
 - CSSI Solar Flux/Space Weather Data
 - Gravitational Survey Data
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -180,27 +182,27 @@ OSTk Physics uses input data from various sources to determine the state of the
...
+| `OSTK_PHYSICS_ENVIRONMENT_MAGNETIC_EARTH_MANAGER_LOCAL_REPOSITORY`                   | `./.open-space-toolkit/physics/data/environment/magnetic/earth`                                        |
+| `OSTK_PHYSICS_ENVIRONMENT_MAGNETIC_EARTH_MANAGER_LOCAL_REPOSITORY_LOCK_TIMEOUT`      | `60`                                                                                                   |
+| `OSTK_PHYSICS_ENVIRONMENT_ATMOSPHERIC_EARTH_MANAGER_MODE`                            | `Automatic`                                                                                            |
+| `OSTK_PHYSICS_ENVIRONMENT_ATMOSPHERIC_EARTH_MANAGER_LOCAL_REPOSITORY`                | `./.open-space-toolkit/physics/data/environment/atmospheric/earth`                                     |
+| `OSTK_PHYSICS_ENVIRONMENT_ATMOSPHERIC_EARTH_MANAGER_LOCAL_REPOSITORY_LOCK_TIMEOUT`   | `60`                                                                                                   |
+| `OSTK_PHYSICS_DATA_REMOTE_URL`                                                       | `https://github.com/open-space-collective/open-space-toolkit-data/raw/v1/data/`                        |
```
kyle-cochran, line 203: Besides whitespace, this is the only line that is changed.

vishwa2710 (approved):
This is awesome.

apaletta3 (approved):
Looks good.

Just to be sure, this doesn't change the fact that none of the ostk-data stuff needs to be installed in a container for ostk to still work properly (the data won't be seeded, but it will still fetch it as needed) and same if the container image doesn't have any of the env var OSTK_PHYSICS_DATA_REPO etc right it still defaults to `./.open-space-toolkit`?

PR #218, opened 2024-03-20
build: fix ostk-core incompatible release

apaletta3:
This should resolve the incompatible versions of ostk-core issue

ostk-physics was letting core go up to 2.2, while io and math were not letting it go past 2.1, that was causing the issue

### DIFF
--- a/bindings/python/requirements.txt
+++ b/bindings/python/requirements.txt
@@ -3,5 +3,6 @@
 numpy>=1.17.4
 
 open-space-toolkit-core~=2.1.0
+open-space-toolkit-core~=2.1.0
 open-space-toolkit-io~=2.0.0
 open-space-toolkit-mathematics~=2.0.0
--- a/bindings/python/requirements.txt
+++ b/bindings/python/requirements.txt
@@ -2,5 +2,5 @@
 
 numpy>=1.17.4
 
-open-space-toolkit-io~=2.0
-open-space-toolkit-mathematics~=2.0
+open-space-toolkit-io~=2.0.0
+open-space-toolkit-mathematics~=2.0.0
--- a/bindings/python/requirements.txt
+++ b/bindings/python/requirements.txt
@@ -3,6 +3,5 @@
 numpy>=1.17.4
 
 open-space-toolkit-core~=2.1.0
-open-space-toolkit-core~=2.1.0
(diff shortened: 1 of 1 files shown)


bindings/python/requirements.txt
```diff
@@ -2,6 +2,5 @@
 
 numpy>=1.17.4
 
-open-space-toolkit-core~=2.1
-open-space-toolkit-io~=2.0
-open-space-toolkit-mathematics~=2.0
+open-space-toolkit-io~=2.0.0
```
kyle-cochran, line 5:
```suggestion
open-space-toolkit-core~=2.1.0
open-space-toolkit-io~=2.0.0
```

I think we should leave the explicit core dependency, as we have it in the cpp

kyle-cochran: Yep!

apaletta3: Wait actually, couldn't we remove the dependency in cpp as well? @kyle-cochran That way we don't have this double dependency?

kyle-cochran: No, unfortunately. Cpp doesn't have a smart dependency resolver like Python does with `pip`. Unless you bake in the required dependency code to your library (this is what compiling _statically_ provides) you have to:
- download
- version control
- link
every single requirement manually. Tools like CMake make this a little easier, but it's still a hassle.

For example, we could statically compile Core. That way, when IO is compiled, all of the Core code is included in the binary, then Physics would have access to it through the IO binary (although linking to the correct function would be a hassle).

Usually you just have shared objects `.so` files. That way both the Physics _and_ IO binaries just use the same one.

apaletta3: Ah ok gotcha, makes sense

bindings/python/requirements.txt
```diff
@@ -2,6 +2,7 @@
 
 numpy>=1.17.4
 
-open-space-toolkit-core~=2.1
-open-space-toolkit-io~=2.0
-open-space-toolkit-mathematics~=2.0
+open-space-toolkit-core~=2.1.0
+open-space-toolkit-core~=2.1.0
```
kyle-cochran, line 6:
```suggestion
```

kyle-cochran (approved):

PR #236, opened 2024-03-28
refactor: simplify bug prone CSSI data fetch

kyle-cochran:
This bug was bubbling up when trying to obtain density measurements for this timestamp, and _specifically_ for the attached CSSI data file:
```
from ostk.physics.time import Scale, Instant, DateTime
from ostk.physics.environment.atmospheric import Earth as EarthAtmosphericModel

from ostk.physics.coordinate import Frame
from ostk.physics.coordinate import Position

atmos = EarthAtmosphericModel(EarthAtmosphericModel.Type.NRLMSISE00)

instant = Instant.date_time(DateTime.parse("2024-03-21 23:00:00.565.999"), Scale.UTC)
position = Position.meters([4029942.66955136, -3891161.1573942, -4022548.89540997], Frame.GCRF())

atmos.get_density_at(position, instant)
```
[SW-Last5Years-breaking.csv](https://github.com/open-space-collective/open-space-toolkit-physics/files/14794326/SW-Last5Years-breaking.csv)

However, this only occurred when that data file was placed into the data repository manually, rather than being manually loaded with the Manager bindings.

This turned out to be a[nother] problem with how the manager tries to auto-load CSSI data files. There was an implicit assumption that data files that were newly loaded from disk would always have observational data for a requested point. 

There's a system that tries to cache the CSSI files that are loaded into memory so that multiple can be referenced at once. This code was mostly leftover cruft from a past era, and for space weather doesn't matter too much because:
- newer files are always preferable, since they have more observational data.
- this system only triggered when either an updated file became available during runtime (happens every 3 days) or we go off the end of the data set (which is 5 years back and 20 years ahead).

I don't think there's any interest in supporting either of these edge cases, since we will hopefully never need a propagation with a runtime of >3 days and will not need to have a single propagation with a span of 25 years. 

This MR simplifies the CSSI fetching part of the code. It should be more bulletproof and slightly faster this way.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth/Manager.cpp
+++ b/bindings/python/src/OpenSpaceToolkitPhysicsPy/Environment/Atmospheric/Earth/Manager.cpp
@@ -15,7 +15,7 @@ inline void OpenSpaceToolkitPhysicsPy_Environment_Atmospheric_Earth_Manager(pybi
         .def("get_mode", &Manager::getMode)
         .def("get_local_repository", &Manager::getLocalRepository)
         .def("get_cssi_space_weather_directory", &Manager::getCSSISpaceWeatherDirectory)
-        .def("get_cssi_space_weather_array", &Manager::getCSSISpaceWeatherArray)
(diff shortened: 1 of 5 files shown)


vishwa2710 (approved):
Oh man, what a fix. <3

PR #249, opened 2024-05-15
feat: remove duration constructor for python, use static methods

vishwa2710:
Amit was burned by this as he didn't realize he was instantiating a Duration with nanoseconds. IMO we should just used the named constructors as it's safest.


bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Duration.cpp
```diff
@@ -22,28 +22,6 @@ inline void OpenSpaceToolkitPhysicsPy_Time_Duration(pybind11::module& aModule)
```
phc1990, line 22: Should we prefix the MR with `!` to signa breaking changes?

phc1990 (approved):
Left nit

PR #250, opened 2024-05-17
feat: add an Instant::Parse utility function and the GPS Epoch

kyle-cochran:
Add a utility function that lets us do:

`Instant.parse("2021-12-02 00:00:00.000", Scale.UTC)`
instead of:
`Instant.date_time(DateTime.parse("2021-12-02 00:00:00.000", Scale.UTC))`

Also adds a shortcut for the GPS epoch, which is Jan 6, 1980 UTC
`Instant.GPS_epoch()`

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Instant.cpp
+++ b/bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Instant.cpp
@@ -196,6 +196,16 @@ inline void OpenSpaceToolkitPhysicsPy_Time_Instant(pybind11::module &aModule)
             )doc"
         )
         .def_static(
+            "GPS_epoch",
+            &Instant::GPSEpoch,
+            R"doc(
+                Get the GPS epoch instant.
+
+                Returns:
+                    Instant: GPS epoch instant.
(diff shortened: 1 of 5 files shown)


bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Instant.cpp
```diff
@@ -235,5 +245,26 @@ inline void OpenSpaceToolkitPhysicsPy_Time_Instant(pybind11::module &aModule)
             )doc"
         )
 
+        .def_static(
+            "parse",
+            &Instant::Parse,
+            arg("string"),
+            arg_v("scale", DEFAULT_TIME_SCALE, "Scale.UTC"),
```
lucas-bremond, line 252: Should we allow an implicit timescale? (it's usually the source of many errors?)

kyle-cochran: Nah good catch. This would be the only constructor with a default Scale... and it's probably not a good idea. Gonna remove.

lucas-bremond (approved):

PR #260, opened 2024-08-01
refactor: loosen datetime conversion limit to 2554 (from 2030)

kyle-cochran:
The hypothesis about this limit is that we want to avoid overflow of the Uint64 nanosecond count value that is stored.
The number of years we can store is roughly 584.8, so this sets the limit to 584 years in the future.

### DIFF
--- a/src/OpenSpaceToolkit/Physics/Time/Instant.cpp
+++ b/src/OpenSpaceToolkit/Physics/Time/Instant.cpp
@@ -432,26 +432,13 @@ Instant Instant::DateTime(const time::DateTime& aDateTime, const Scale& aTimeSca
         throw ostk::core::error::runtime::Wrong("Scale");
     }
 
-    if ((aDateTime.accessDate().getYear() < 1970) || (aDateTime.accessDate().getYear() > 2030))
+    // Limit the year based on the max number of nanoseconds we can store with a UInt64
+    if ((aDateTime.accessDate().getYear() < 1970) || (aDateTime.accessDate().getYear() > 2500))
     {
         throw ostk::core::error::RuntimeError(
-            "DateTime year {} out of supported range [{} - {}]", aDateTime.accessDate().getYear(), 1970, 2030
+            "DateTime year {} out of supported range [{} - {}]", aDateTime.accessDate().getYear(), 1970, 2500
         );
     }
-
-    // auto getTimePointString =
-    // [ ] (const std::chrono::time_point<std::chrono::system_clock>& aTimePoint) -> String
-    // {
-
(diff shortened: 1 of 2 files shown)


vishwa2710 (approved):
Looks good! Just have to fix python tests

src/OpenSpaceToolkit/Physics/Time/Instant.cpp
```diff
@@ -433,10 +433,10 @@ Instant Instant::DateTime(const time::DateTime& aDateTime, const Scale& aTimeSca
     }
 
     // Limit the year based on the max number of nanoseconds we can store with a UInt64
-    if ((aDateTime.accessDate().getYear() < 1970) || (aDateTime.accessDate().getYear() > 2500))
```
lucas-bremond: Why was it previously `2030`? (wasn't the previous value `2500`?)

test/OpenSpaceToolkit/Physics/Time/Instant.test.cpp
```diff
@@ -1777,7 +1777,7 @@ TEST(OpenSpaceToolkit_Physics_Time_Instant, DateTime)
         for (const auto scale : scales)
         {
             EXPECT_ANY_THROW(Instant::DateTime(DateTime(1969, 1, 1, 0, 0, 0), scale));
-            EXPECT_ANY_THROW(Instant::DateTime(DateTime(2031, 1, 1, 0, 0, 0), scale));
+            EXPECT_ANY_THROW(Instant::DateTime(DateTime(2501, 1, 1, 0, 0, 0), scale));
```
vishwa2710, line 1780:
```suggestion
            EXPECT_ANY_THROW(Instant::DateTime(DateTime(2555, 1, 1, 0, 0, 0), scale));
```

vishwa2710 (approved):

src/OpenSpaceToolkit/Physics/Time/Instant.cpp
vishwa2710: The previous value was 2030. I don't recall it being 2500. Looking at the git blame, it hasn't been changed in 6 years.

lucas-bremond: Looks like a reviewed a commit diff by mistake, not the PR diff. And one commit went from 2500 to 2554, hence the confusion.

vishwa2710: Ah gotcha. Yeah Kyle changed it from 2500 to 2554 to match exactly the number of nanoseconds an Uint64 can track when converted to years (rounded down).

PR #267, opened 2024-09-19
feat: global environment singleton

vishwa2710:
In this MR, we add the concept of a `central celestial object` to the Environment, and additionally create a "global" environment instance which is a singleton. This is to simplify the instantiation of several objects in OSTk Physics, Astro and Simulation which typically operate for a central celestial object.

Like in simulation toolkits, where you specify your environment and then conduct your analysis, we can now create an Environment and set it to global. This environment will then be used as required by various functions.
We still retain the ability to provide celestial specific parameters, if we want to override the global environment.
Therefore this is a backwards compatible change.

```
from ostk.physics import Environment
from ostk.physics.environment.object.celestial import Earth
from ostk.physics.coordinate.spherical import LLA

environment = Environment.default(set_global_instance=True)

lla_1 = LLA.vector([0.0, 0.0, 0.0])
lla_2 = LLA.vector([1.0, 0.0, 0.0])

distance = lla_1.calculate_distance_to(lla_2)  # 110574.38855779878 [m]

egm96_earth = Earth.EGM96()
eg96_distance = lla_1.calculate_distance_to(
    lla_2,
    egm96_earth.get_equatorial_radius(),
    egm96_earth.get_flattening()
) # 110574.37642226892 [m]
```
We can instantiate orbits similarly and convert COEs etc. using these defaults.


include/OpenSpaceToolkit/Physics/Environment.hpp
```diff
@@ -73,6 +86,12 @@ class Environment
 
     bool hasObjectWithName(const String& aName) const;
 
+    /// @brief              Has central celestial
+    ///
+    /// @return             True if environment has central celestial
+
+    bool hasCentralCelestial() const;
```
lucas-bremond, line 93:
```suggestion
    bool hasCentralObject() const;
```

to be in semantic sync w/ the rest?

include/OpenSpaceToolkit/Physics/Environment.hpp
```diff
@@ -39,6 +41,17 @@ class Environment
 
     Environment(const Instant& anInstant, const Array<Shared<Object>>& anObjectArray);
 
+    /// @brief              Constructor
+    /// @param              [in] aCentralBody A central body
+    /// @param              [in] anObjectArray An array of shared pointers to objects
+    /// @param              [in] anInstant An instant. Defaults to J2000 epoch.
+
+    Environment(
+        const Shared<Object>& aCentralBody,
```
lucas-bremond, line 50:
```suggestion
        const Shared<Celestial>& aCentralCelestialObject,
```

no?

include/OpenSpaceToolkit/Physics/Environment.hpp
```diff
@@ -104,6 +123,12 @@ class Environment
 
     Shared<const Celestial> accessCelestialObjectWithName(const String& aName) const;
 
+    /// @brief              Access central celestial
+    ///
+    /// @return             Shared pointer to central celestial
+
+    Shared<const Celestial> accessCentralCelestial() const;
```
lucas-bremond, line 130:
```suggestion
    Shared<const Celestial> accessCentralCelestialObject() const;
```

include/OpenSpaceToolkit/Physics/Environment.hpp
```diff
@@ -116,6 +141,10 @@ class Environment
 
     Array<String> getObjectNames() const;
 
+    /// @brief              Get central body
+    ///
+    /// @return             Shared pointer to central body
+
```
lucas-bremond, line 147: Missing getter

include/OpenSpaceToolkit/Physics/Environment.hpp
```diff
@@ -161,16 +190,34 @@ class Environment
...
     /// @return             Undefined environment
 
-    static Environment Default();
+    static Environment Default(const bool& setGlobalInstance = false);
+
+    /// @brief              Get the singleton instance of the environment
+    ///
+    /// @return             Shared pointer to the singleton instance
+
+    static Shared<Environment> AccessInstance();
```
lucas-bremond, line 206: Clearer?

```suggestion
    static Shared<Environment> AccessGlobalInstance();
```

include/OpenSpaceToolkit/Physics/Environment.hpp
```diff
@@ -161,16 +190,34 @@ class Environment
...
+
+    static void SetInstance(const Shared<Environment>& instance);
 
    private:
     Instant instant_;
     Array<Shared<const Object>> objects_;
+    Shared<const Object> centralCelestial_;
+
+    static std::shared_mutex mutex_;
+    static Shared<Environment> instance_;
```
lucas-bremond, line 220: Should these static properties should be part of the object itself? I think they should better live outside no?

include/OpenSpaceToolkit/Physics/Environment.hpp
```diff
@@ -161,16 +190,34 @@ class Environment
...
+    ///
+    /// @return             Shared pointer to the singleton instance
+
+    static Shared<Environment> AccessInstance();
+
+    /// @brief              Set the singleton instance of the environment
+    ///
+    /// @param              [in] instance Shared pointer to the new singleton instance
+
+    static void SetInstance(const Shared<Environment>& instance);
```
lucas-bremond, line 212: Do we want to allow this to be changed externally once set?

I'd perhaps hold on that one for now (it's easy to add later)

include/OpenSpaceToolkit/Physics/Coordinate/Spherical/LLA.hpp
```diff
@@ -40,20 +40,58 @@ using ostk::physics::unit::Angle;
 class LLA
 {
    public:
+    /// @brief Constructs an LLA (Latitude, Longitude, Altitude) object.
```
lucas-bremond, line 43: Thanks for being such a good citizen

include/OpenSpaceToolkit/Physics/Coordinate/Spherical/LLA.hpp
```diff
@@ -67,6 +105,13 @@ class LLA
         const LLA& aLLA, const Length& anEllipsoidEquatorialRadius, const Real& anEllipsoidFlattening
     ) const;
 
+    /// @brief                  Calculate the distance between this LLA coordinate and another LLA coordinate
+    ///
+    /// @param                  [in] aLLA A LLA coordinate
+    /// @return
+
+    Length calculateDistanceTo(const LLA& aLLA) const;
```
lucas-bremond, line 113: I think we should stick to a rule that we always expose the explicit variant too? (if we use default args, then my point is moot)

include/OpenSpaceToolkit/Physics/Environment.hpp
vishwa2710: Hmmmmm I'm a bit confused. How would a user set it if we don't provide a setter?

src/OpenSpaceToolkit/Physics/Environment.cpp
```diff
@@ -28,16 +34,39 @@ Environment::Environment(const Instant& anInstant, const Array<Shared<Object>>&
...
+{
+    objects_.reserve(anObjectArray.getSize() + 1);
+
+    const Shared<const Object> centralBody = Shared<const Object>(aCentralBody->clone());
+    centralCelestial_ = centralBody;
+
+    objects_.add(centralBody);
+    for (const auto& objectSPtr : anObjectArray)
+    {
+        objects_.add(Shared<const Object>(objectSPtr->clone()));
```
phc1990, line 52: should we check and raise an error if `centralBody` is also part of the `anObjectArray` ? (to avoid adding it twice)

include/OpenSpaceToolkit/Physics/Environment.hpp
phc1990: > How would a user set it if we don't provide a setter?

Maybe we can just add the `setToGlobal` to the actual arguments Environment Constructor too (defaulting to false). So you can create any Environment you want and set it to global on the spot:

```python
my_environment = Environment(my_objects, set_to_global=True)
```

src/OpenSpaceToolkit/Physics/Coordinate/Spherical/LLA.cpp
```diff
@@ -109,6 +112,20 @@ Length LLA::calculateDistanceTo(
...
+    {
+        throw ostk::core::error::runtime::Undefined("Central celestial");
+    }
+
+    return this->calculateDistanceTo(
+        aLLA,
+        Environment::AccessInstance()->accessCentralCelestial()->getEquatorialRadius(),
+        Environment::AccessInstance()->accessCentralCelestial()->getFlattening()
+    );
+}
```
phc1990, line 127: Maybe I'm lacking contest here, my only feedback is that the fact that this is measured w.r.t. to the global Environment's central body is a bit hidden. Maybe we can allow for an **optional ** Central Body or Environment argument:

```suggestion
Length LLA::calculateDistanceTo(const LLA& aLLA, const Shared<const Celestial> aCentralBody) const
{
    // If no Central Body default to global Environment central boody
    
    return this->calculateDistanceTo(
        aLLA,
        aCentralBody->getEquatorialRadius(),
        aCentralBody->getFlattening()
    );
}
```

Or 

```suggestion
Length LLA::calculateDistanceTo(const LLA& aLLA, const Shared<const Environment> anEnvironment) const
{
    // If no environment, default to global Environment

    return this->calculateDistanceTo(
        aLLA,
        anEnvironment->accessCentralCelestial()->getEquatorialRadius(),
        anEnvironment->accessCentralCelestial()->getFlattening()
    );
}
```

phc1990 (commented):
Left some nit comments, nice stuff!

src/OpenSpaceToolkit/Physics/Coordinate/Spherical/LLA.cpp
vishwa2710: I agree it's not obvious. I think we should update the OSTk README and explicitly explain how the new global environment singleton will be used across the repos. I will push a commit let me know.

src/OpenSpaceToolkit/Physics/Environment.cpp
vishwa2710: In this case we actually want to store a pointer to the central object so we can easily access it. Otherwise everywhere we would have to loop through the objects + central body.
In this case, we store a shared pointer to the central body so it's lightweight.

lucas-bremond (approved):

PR #273, opened 2024-10-02
fix: manual manager mode file loading

vishwa2710:
The desired behaviour is:
- Automatic: Load **and** fetch files
- Manual: Load files


src/OpenSpaceToolkit/Physics/Coordinate/Frame/Provider/IERS/Manager.cpp
```diff
@@ -514,11 +514,25 @@ const BulletinA* Manager::accessBulletinA_() const
         return &bulletinA_;
     }
 
-    // No bulletin A found and unable to fetch
-
+    if (mode_ == Manager::Mode::Manual)
```
lucas-bremond, line 517: Should we use a switch statement?

phc1990 (approved):

PR #278, opened 2024-10-04
feat: remove clone of const objects in Environment

vishwa2710:
## Summary by CodeRabbit

- **New Features**
	- Simplified construction of `Environment` objects by directly initializing with provided arrays and objects, improving performance.
  
- **Bug Fixes**
	- Removed unnecessary cloning operations to streamline memory management.

- **Refactor**
	- Eliminated copy constructor and assignment operator to enhance design efficiency and reduce complexity.

### DIFF
--- a/include/OpenSpaceToolkit/Physics/Environment.hpp
+++ b/include/OpenSpaceToolkit/Physics/Environment.hpp
@@ -57,19 +57,6 @@ class Environment
         const bool& setGlobalInstance = false
     );
 
-    /// @brief              Copy constructor
-    ///
-    /// @param              [in] anEnvironment An environment
-
-    Environment(const Environment& anEnvironment);
-
-    /// @brief              Copy assignment operator
-    ///
-    /// @param              [in] anEnvironment An environment
-    /// @return             Reference to environment
-
-    Environment& operator=(const Environment& anEnvironment);
-
     /// @brief              Output stream operator
     ///
     /// @param              [in] anOutputStream An output stream
--- a/src/OpenSpaceToolkit/Physics/Environment.cpp
+++ b/src/OpenSpaceToolkit/Physics/Environment.cpp
@@ -26,16 +26,9 @@ Environment::Environment(
     const Instant& anInstant, const Array<Shared<const Object>>& anObjectArray, const bool& setGlobalInstance
 )
     : instant_(anInstant),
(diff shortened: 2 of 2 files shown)


src/OpenSpaceToolkit/Physics/Environment.cpp
```diff
@@ -50,17 +43,14 @@ Environment::Environment(
 )
     : instant_(anInstant),
       objects_(Array<Shared<const Object>>::Empty()),
-      centralCelestialObject_(nullptr)
+      centralCelestialObject_(aCentralCelestialObject)
```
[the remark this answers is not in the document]
vishwa2710: This is not an issue as it is a shared pointer to a `const Object`.

src/OpenSpaceToolkit/Physics/Environment.cpp
```diff
@@ -50,17 +43,14 @@ Environment::Environment(
...
 
-    const Shared<const Object> centralCelestialObject = Shared<const Object>(aCentralCelestialObject->clone());
-    centralCelestialObject_ = centralCelestialObject;
-
-    objects_.add(centralCelestialObject);
+    objects_.add(centralCelestialObject_);
     for (const auto& objectSPtr : anObjectArray)
     {
-        objects_.add(Shared<const Object>(objectSPtr->clone()));
+        objects_.add(objectSPtr);
```
lucas-bremond, line 53: This is risky because it means you're sharing a pointer supplied from the outside, no there's no clear ownership anymore and subtle side effects could occur (what if these object points came from another Environment instance, that is still used in parallel to this one?)

include/OpenSpaceToolkit/Physics/Environment.hpp
```diff
@@ -57,19 +57,6 @@ class Environment
         const bool& setGlobalInstance = false
     );
 
-    /// @brief              Copy constructor
-    ///
-    /// @param              [in] anEnvironment An environment
-
-    Environment(const Environment& anEnvironment);
```
lucas-bremond, line 64: Do you intent to use the defaults version, or forbid their use?

vishwa2710: Intend to use the default version.

PR #279, opened 2024-10-14
fix: Engine manual mode load kernels

vishwa2710:
Updated the Engine to load kernels in both manual and automatic mode. In manual mode, if the kernel does not exist, it raises an error, in automatic mode, the kernel is fetched and then loaded.

## Summary by CodeRabbit

- **New Features**
	- Enhanced kernel loading process for improved flexibility.
	- Introduced a comprehensive suite of unit tests for the `Engine` class, validating its functionality.

- **Bug Fixes**
	- Updated expected behavior in tests to reflect changes in the `getOriginIn` function.

- **Documentation**
	- Improved clarity and accessibility of types in test files.

### DIFF
--- a/src/OpenSpaceToolkit/Physics/Environment/Ephemeris/SPICE/Engine.cpp
+++ b/src/OpenSpaceToolkit/Physics/Environment/Ephemeris/SPICE/Engine.cpp
@@ -351,20 +351,11 @@ void Engine::setup()
 
     errdev_c("SET", 4096, (SpiceChar*)"NULL");
 
-    if (mode_ == Engine::Mode::Automatic)
+    // Load default kernels
+
+    for (const auto& kernel : Engine::DefaultKernels())
     {
-        Directory kernelDirectory = Manager::Get().getLocalRepository();
-        if (!kernelDirectory.exists())
-        {
-            kernelDirectory.create();
-        }
-
-        // Load default kernels
-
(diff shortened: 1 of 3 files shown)


src/OpenSpaceToolkit/Physics/Environment/Ephemeris/SPICE/Engine.cpp
```diff
@@ -351,20 +351,11 @@ void Engine::setup()
...
-    if (mode_ == Engine::Mode::Automatic)
-    {
-        Directory kernelDirectory = Manager::Get().getLocalRepository();
-        if (!kernelDirectory.exists())
-        {
-            kernelDirectory.create();
-        }
-
-        // Load default kernels
+    // Load default kernels
```
apaletta3, line 354: You don't need to lock a mutex here? I see it being used above before calling the `this->loadKernel_(kernel);` function

apaletta3 (approved):
LGTM! One comment

src/OpenSpaceToolkit/Physics/Environment/Ephemeris/SPICE/Engine.cpp
vishwa2710: The setup function is called in the constructor at which point no mutex is required I believe as there is only a single instance of the Engine. And the second way the setup function is called is in reset which has it's own lock guard.

PR #298, opened 2024-12-12
feat: add J3 term to model parameters

alex-liang3:
Adds an optional C30/J3 Parameter to the Gravitational Model, and populated them for EGM2008 and WGS84/EGM96. To maintain backwards-compatibility, I've chosen to default these to 0 rather than `Undefined` - if they were `Undefined`, an existing `Parameters` struct without C30/J3 would suddenly stop being `isDefined()`, which I don't think is a desirable behaviour. Mathematically, I think this is consistent too. 

The values for EGM2008 were manually parsed from the coefficients file provided by GeographicLib, and cross-referenced with Vallado's Fundamentals of Astrodynamics and Applications (Table D-1). The values for WGS84/EGM96 came from the official NIMA spec (Table 5.1).
[wgs84fin.pdf](https://github.com/user-attachments/files/18115017/wgs84fin.pdf)

## Summary by CodeRabbit

- **New Features**
	- Introduced additional parameters `C30` and `J3` to the gravitational model for enhanced precision.
	- Updated constructors to accommodate new parameters in gravitational modeling.

- **Bug Fixes**
	- Improved validation and assertions in tests for gravitational model parameters.

- **Tests**
	- Expanded test coverage for gravitational models to include new scenarios and assertions for `C30` and `J3` coefficients.
	- Added a new test case to verify output streaming for the gravitational model.


vishwa2710 (approved):
Yeah I agree this is less than ideal, but it allows us to maintain backwards compatibility so I think this is fine for now.
I would say let's put a comment stating that we will change this in a future major release and make the J3 term required.
This forces the user to provide a dummy default value.

vishwa2710:
Looks like we're missing a test for the ostream friend function, if you could add one it would be great. Should be fairly straightforward you can copy the pattern from any other test file.

src/OpenSpaceToolkit/Physics/Environment/Gravitational/Earth.cpp
```diff
@@ -45,6 +46,7 @@ const Model::Parameters Earth::WGS84_EGM96 = {
     Length::Meters(6378137.0),
     1.0 / 298.257223563,
     -4.841653717360e-04,
+    9.57254173792e-07,
```
apaletta3, line 49: I think you should also add it to the EGM96 model, since that's the one we use most often - we almost never use the WGS84_EGM96 model since its a weird hybrid.

PR #326, opened 2025-02-20
fix: Interval binding docstrings

apaletta3:
While using the Interval code, I noticed that the chat-gpt generated docstrings didn't match the stub types so I fixed the docstrings for the whole Interval class in the bindings file.

Also, there were several instances of missing kwargs, so for now I just removed the kwarg from the docstring so it is clear to the user that they can't use a kwarg as an argument in that function, only a positional arg. I thought of adding kwargs where they were missing, but then I realized that this might be an ABI break, and so its not worth it to do this for such a small fix right now.

In the future, we will have to address these discrepancies and fix them

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Interval.cpp
+++ b/bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Interval.cpp
@@ -191,7 +191,7 @@ inline void OpenSpaceToolkitPhysicsPy_Time_Interval(pybind11::module& aModule)
                 Get the intersection with another interval.
 
                 Args:
-                    interval (Interval): Another interval.
+                    (Interval): Another interval.
 
                 Returns:
                     Interval: Intersection interval (Undefined if there is no intersection).
@@ -204,7 +204,7 @@ inline void OpenSpaceToolkitPhysicsPy_Time_Interval(pybind11::module& aModule)
                 Get the union with another interval.
 
                 Args:
-                    interval (Interval): Another interval.
+                    (Interval): Another interval.
 
                 Returns:
                     Interval: Union interval (Undefined if there is no single-interval union).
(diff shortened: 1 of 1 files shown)


bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Interval.cpp
```diff
@@ -191,7 +191,7 @@ inline void OpenSpaceToolkitPhysicsPy_Time_Interval(pybind11::module& aModule)
                 Get the intersection with another interval.
 
                 Args:
-                    interval (Interval): Another interval.
+                    (Interval): Another interval.
```
vishwa2710, line 194: I think you misunderstood the format for the Args field, you have to indeed provide the name of the argument, the type and then the description. This is correct IMO.

bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Interval.cpp
```diff
@@ -231,10 +231,10 @@ inline void OpenSpaceToolkitPhysicsPy_Time_Interval(pybind11::module& aModule)
                 Generate a grid of instants with a given time step.
 
                 Args:
-                    aTimeStep (Duration): Time step.
```
vishwa2710, line 234: This should be `time_step`

vishwa2710 (approved):

bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Interval.cpp
apaletta3: discusssed that this is because there is no kwarg here

bindings/python/src/OpenSpaceToolkitPhysicsPy/Time/Interval.cpp
apaletta3: discusssed that this is because there is no kwarg here

PR #329, opened 2025-02-24
fix: update default earth to EGM2008 20x20

vishwa2710:
## Summary by CodeRabbit

- **New Features**
  - The Earth simulation now employs a refined gravitational configuration to deliver enhanced accuracy and consistency in behavior.

- **Chores**
  - Internal initialization processes have been streamlined, resulting in improved maintainability and performance.

### DIFF
--- a/src/OpenSpaceToolkit/Physics/Environment/Object/Celestial/Earth.cpp
+++ b/src/OpenSpaceToolkit/Physics/Environment/Object/Celestial/Earth.cpp
@@ -126,7 +126,7 @@ Earth Earth::FromModels(
 
 Earth Earth::Default()
 {
-    return Earth::EGM2008();
+    return Earth::EGM96(20, 20);
 }
 
 Earth Earth::EGM2008(const Integer& aGravityModelDegree, const Integer& aGravityModelOrder)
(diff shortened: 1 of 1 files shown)


vishwa2710:
> Should we mark this as a breaking change?

I was thinking about it, since we're only updating the gravity model order and degrees I think it's fine. If the behaviour was more significant I would. Also IMO this is more of a fix than it is a behavioural change as a full gravity order potential computation is ridiculous...

src/OpenSpaceToolkit/Physics/Environment/Object/Celestial/Earth.cpp
```diff
@@ -145,8 +145,6 @@ Earth Earth::WGS84_EGM96(const Integer& aGravityModelDegree, const Integer& aGra
 
 Earth Earth::EGM96(const Integer& aGravityModelDegree, const Integer& aGravityModelOrder)
 {
-    const Shared<const Frame> earthFrameSPtr = Frame::ITRF();
```
alex-liang3, line 148: Looks like this might be causing the tests to fail... if we need to leave it in, would be good to leave a comment explaining why it's necessary

alex-liang3 (approved):

src/OpenSpaceToolkit/Physics/Environment/Object/Celestial/Earth.cpp
vishwa2710: Looking good now.

PR #330, opened 2025-02-24
feat: add interval get_type method

apaletta3:


src/OpenSpaceToolkit/Physics/Time/Interval.cpp
```diff
@@ -58,6 +58,16 @@ const Instant& Interval::accessEnd() const
...
 }
 
+Interval::Type Interval::getType() const
+{
+    if (!this->isDefined())
+    {
+        throw ostk::core::error::runtime::Undefined("Interval");
+    }
+
+    return this->getType();
```
apaletta3, line 68:
```suggestion
    return mathematics::object::Interval<Instant>::getType();
```

src/OpenSpaceToolkit/Physics/Time/Interval.cpp
```diff
@@ -58,6 +58,16 @@ const Instant& Interval::accessEnd() const
     return this->accessUpperBound();
 }
 
+Interval::Type Interval::getType() const
```
alex-liang3, line 61: This is actually already available in cpp since this Time Interval inherits from the Interval in ostk-physics.

alex-liang3 (commented):
Done [here](https://github.com/open-space-collective/open-space-toolkit-physics/pull/340)

PR #340, opened 2025-07-07
feat: add get_type to Interval

alex-liang3:
Adds a python binding to get the Type of an Interval (e.g. Open, Closed, HalfOpenLeft, HalfOpenRight) 

## Summary by CodeRabbit

* **New Features**
  * Added a method to retrieve the type of an interval in the Python interface.

* **Tests**
  * Added a test to verify the interval type retrieval method.


apaletta3 (approved):
Just double check that the binding doesn't come for free with the inheritance from the ostk-math type

PR #343, opened 2025-08-05
fix: duplicate celestial objects in environment

apaletta3:
Realized that I had accidentally created an environment with two suns the other day, so I think we should explicitly disallow that. The logic has to check for similarity based on the name of the object because there's no way to downcast a `Celestial` to all child types because users could have extended ostk classes themselves with their own `Celestial` implementations.

Therefore if someone still wants to (for whatever reason) create two earths, they could still do it by naming the second Earth something else.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented creating environments with duplicate celestial object names by adding validation that raises an error on duplicates.
* **Documentation**
  * Clarified in user-facing docs and Python bindings that constructors will raise an error if duplicate celestial object names are provided.
* **Tests**
  * Added tests verifying an error is raised when constructing an environment with duplicate celestial object names.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/OpenSpaceToolkit/Physics/Environment.cpp
+++ b/src/OpenSpaceToolkit/Physics/Environment.cpp
@@ -30,6 +30,8 @@ Environment::Environment(
       objects_(anObjectArray),
       centralCelestialObject_(nullptr)
 {
+    this->validateCelestialObjects();
+
     if (setGlobalInstance)
     {
         Environment::SetGlobalInstance(std::make_shared<Environment>(*this));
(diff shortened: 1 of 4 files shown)


src/OpenSpaceToolkit/Physics/Environment.cpp
```diff
@@ -375,5 +379,28 @@ bool Environment::HasGlobalInstance()
     return instance != nullptr;
 }
 
+void Environment::validateCelestialObjects()
+{
+    Array<String> celestialNames = Array<String>::Empty();
```
vishwa2710:
```suggestion
    Array<String> celestialNames = Array<String>::Empty();
```
Nit: More performant to use an unordered_set here

PR #344, opened 2025-08-06
fix: environment unsafe central object behavior

apaletta3:
We shouldn't be using a `static_pointer_cast`, because they are only for upcasting or safe downcasting where you're sure that the return type is the derived class type. Since there is no guarantee that the `centralCelestialObjectSPtr_` is of type `Shared<const Celestial>` instead of type `Shared<const Object>`, we should use `dynamic_pointer_cast` instead and throw and error if it returns a nullpointer.

This technically would change the behavior of the method in question if someone has an `Environment` with a non-Celestial `Object` as the `centralCelestialObjectSPtr_`.

Source: https://stackoverflow.com/questions/28002/regular-cast-vs-static-cast-vs-dynamic-cast

## Summary by CodeRabbit

* **Refactor**
  * Improved internal handling of environment objects and central celestial object to enhance reliability and error detection, with no impact on user-facing features.

### DIFF
--- a/src/OpenSpaceToolkit/Physics/Environment.cpp
+++ b/src/OpenSpaceToolkit/Physics/Environment.cpp
@@ -206,7 +206,7 @@ Shared<const Celestial> Environment::accessCentralCelestialObject() const
 
     if (const auto centralCelestialObjectSPtr = std::dynamic_pointer_cast<const Celestial>(centralCelestialObject_))
     {
-        return celestialObjectSPtr;
+        return centralCelestialObjectSPtr;
     }
     
     throw ostk::core::error::RuntimeError("No central celestial object.");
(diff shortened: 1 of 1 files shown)


src/OpenSpaceToolkit/Physics/Environment.cpp
```diff
@@ -213,12 +204,14 @@ Shared<const Celestial> Environment::accessCentralCelestialObject() const
         throw ostk::core::error::runtime::Undefined("Environment");
     }
 
-    if (centralCelestialObject_ == nullptr)
+    if (const auto centralCelestialObjectSPtr = std::dynamic_pointer_cast<const Celestial>(centralCelestialObject_))
     {
-        throw ostk::core::error::RuntimeError("No central celestial object.");
+        return celestialObjectSPtr;
```
vishwa2710, line 209:
```suggestion
        return centralCelestialObjectSPtr;
```

PR #348, opened 2025-10-01
fix: space weather fetching

vishwa2710:
This MR fixes a few different issues that were found in the Manifest manager as well as the space weather data manager.

Problem:
  - Manifest Manager's manifest refresh had inverted logic that prevented automatic updates (comparing nextUpdateCheckTimestamp < lastModifiedTimestamp instead of Instant::Now() > nextUpdateCheckTimestamp)
  - Atmospheric Manager didn't check if remote space weather files were newer
    than local copies, leading to stale data being used indefinitely

  Solution:
  - Fixed manifest refresh condition to check if current time exceeds next
    update check OR if manifest age exceeds refresh rate (24h default)
  - Added lastModifiedTimestamp tracking to CSSISpaceWeather class
  - Updated Atmospheric Manager to follow IERS pattern: load local file, query
    manifest for remote timestamp, and fetch if remote is newer
  - Added tests for both manifest refresh conditions and timestamp checking

### DIFF
--- a/src/OpenSpaceToolkit/Physics/Data/Manager.cpp
+++ b/src/OpenSpaceToolkit/Physics/Data/Manager.cpp
@@ -180,8 +180,8 @@ void Manager::checkManifestAgeAndUpdate_() const
 
     const Duration manifestAge = Instant::Now() - manifest_.getLastModifiedTimestamp();
 
-    // If loaded manifest is old enough, fetch a new one and load it.
-    if (nextUpdateCheckTimestamp < manifest_.getLastModifiedTimestamp() && manifestAge > DataRefreshRate_())
+    // If current time is past next update check time or manifest is old enough, fetch a new one and load it.
+    if (Instant::Now() > nextUpdateCheckTimestamp || manifestAge > DataRefreshRate_())
     {
         if (mode_ == Manager::Mode::Manual)
         {
--- a/src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Manager.cpp
+++ b/src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Manager.cpp
@@ -373,16 +373,39 @@ const CSSISpaceWeather* Manager::accessCSSISpaceWeatherAt_(const Instant& anInst
 
 File Manager::getLatestCSSISpaceWeatherFile_() const
 {
-    // Parse CSSI Space Weather Directories, e.g.,
-    // `.open-space-toolkit/physics/environment/atmospheric/earth/CSSI-Space-Weather/2022-05-19/`, and find the
-    // latest one.
+    File localCSSISpaceWeatherFile = File::Undefined();
 
+    // Check if local file exists
     if (this->getCSSISpaceWeatherDirectory().containsFileWithName(CSSISpaceWeatherFileName))
     {
-        return File::Path(this->getCSSISpaceWeatherDirectory().getPath() + Path::Parse(CSSISpaceWeatherFileName));
+        localCSSISpaceWeatherFile =
+            File::Path(this->getCSSISpaceWeatherDirectory().getPath() + Path::Parse(CSSISpaceWeatherFileName));
+
+        // Load local file to access its timestamp
+        const CSSISpaceWeather localSpaceWeather = CSSISpaceWeather::Load(localCSSISpaceWeatherFile);
+
+        // Get the Data Manager instance to query manifest
+        ManifestManager& manifestManager = ManifestManager::Get();
+
+        // Query when the remote file was last updated (this may trigger a manifest fetch)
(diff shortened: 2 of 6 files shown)


src/OpenSpaceToolkit/Physics/Data/Manager.cpp
```diff
@@ -180,8 +180,8 @@ void Manager::checkManifestAgeAndUpdate_() const
 
     const Duration manifestAge = Instant::Now() - manifest_.getLastModifiedTimestamp();
 
-    // If loaded manifest is old enough, fetch a new one and load it.
-    if (nextUpdateCheckTimestamp < manifest_.getLastModifiedTimestamp() && manifestAge > DataRefreshRate_())
+    // If current time is past next update check time or manifest is old enough, fetch a new one and load it.
+    if (Instant::Now() > nextUpdateCheckTimestamp || manifestAge > DataRefreshRate_())
```
alex-liang3, line 184: Should this remain an `AND` condition? so that `DataRefreshRate_()` remains as a "maximum" frequency. Also, this means we won't unnecessarily check this every `DataRefreshRate_()`, when we haven't yet reached the `nextUpdateCheckTimestamp` (and therefore we "know" there isn't any new data to fetch)

```suggestion
    if (Instant::Now() > nextUpdateCheckTimestamp && manifestAge > DataRefreshRate_())
```

test/OpenSpaceToolkit/Physics/Data/Manager.test.cpp
```diff
@@ -176,6 +181,146 @@ TEST_F(OpenSpaceToolkit_Physics_Data_Manager, GetLastUpdateTimestampFor_Skipped)
     }
 }
 
+TEST_F(OpenSpaceToolkit_Physics_Data_Manager, ManifestRefreshWhenPastNextUpdateCheckTime)
+{
+    // Create a test manifest with next_update_check in the past
+    const String testManifestContent = R"({
```
alex-liang3, line 187: there's already a `manifest.json` in the test files - doesn't that work here too?

test/OpenSpaceToolkit/Physics/Data/Manager.test.cpp
```diff
@@ -176,6 +181,146 @@ TEST_F(OpenSpaceToolkit_Physics_Data_Manager, GetLastUpdateTimestampFor_Skipped)
...
+            }
+        },
+        ostk::core::error::RuntimeError
+    );
+
+    // Cleanup
+    tempDir.remove();
+}
+
+TEST_F(OpenSpaceToolkit_Physics_Data_Manager, ManifestRefreshWhenAgeExceedsRefreshRate)
```
alex-liang3, line 248: see https://github.com/open-space-collective/open-space-toolkit-physics/pull/348/files#r2397984311 - i'm not sure this is actually desirable behaviour

src/OpenSpaceToolkit/Physics/Environment/Atmospheric/Earth/Manager.cpp
```diff
@@ -373,16 +373,39 @@ const CSSISpaceWeather* Manager::accessCSSISpaceWeatherAt_(const Instant& anInst
...
+        }
+    }
+    else
+    {
+        // No local file exists, fetch from remote
+        localCSSISpaceWeatherFile = const_cast<Manager*>(this)->fetchLatestCSSISpaceWeather_();
     }
 
-    return const_cast<Manager*>(this)->fetchLatestCSSISpaceWeather_();
+    return localCSSISpaceWeatherFile;
```
alex-liang3, line 408: nit: early returns reads cleaner to me, but up to you.

```suggestion
    // If local file doesn't exist, fetch from remote.
    if (!this->getCSSISpaceWeatherDirectory().containsFileWithName(CSSISpaceWeatherFileName))
    {
        return const_cast<Manager*>(this)->fetchLatestCSSISpaceWeather_();
    }

    File localCSSISpaceWeatherFile =
        File::Path(this->getCSSISpaceWeatherDirectory().getPath() + Path::Parse(CSSISpaceWeatherFileName));

    // Load local file to access its timestamp
    const CSSISpaceWeather localSpaceWeather = CSSISpaceWeather::Load(localCSSISpaceWeatherFile);

    // Get the Data Manager instance to query manifest
    ManifestManager& manifestManager = ManifestManager::Get();

    // Query when the remote file was last updated (this may trigger a manifest fetch)
    const Instant remoteUpdateTimestamp = manifestManager.getLastUpdateTimestampFor(CSSISpaceWeatherManifestName);

    // Get when the local file was last modified
    const Instant localUpdateTimestamp = localSpaceWeather.accessLastModifiedTimestamp();

    // If remote is newer, fetch it
    if (remoteUpdateTimestamp > localUpdateTimestamp)
    {
        return const_cast<Manager*>(this)->fetchLatestCSSISpaceWeather_();
    }

    return localCSSISpaceWeatherFile;
```

alex-liang3 (approved):

PR #349, opened 2025-10-27
feat!: add overwrite option to Frame construction method

loft-nicolas-sanchez:
## Summary by CodeRabbit

* **New Features**
  * Added an `overwrite` option to Frame construction (including the Python binding) to optionally replace an existing frame with the same name.

* **Behavior**
  * When `overwrite=false`, creating a Frame with a duplicate name raises an error.
  * When `overwrite=true`, the existing frame is replaced by the newly constructed one.

* **Tests**
  * Added unit tests covering duplicate-name failures and successful replacement when overwrite is enabled.


bindings/python/src/OpenSpaceToolkitPhysicsPy/Coordinate/Frame.cpp
```diff
@@ -356,6 +357,7 @@ inline void OpenSpaceToolkitPhysicsPy_Coordinate_Frame(pybind11::module& aModule
                     is_quasi_inertial (bool): True if quasi-inertial.
                     parent_frame (Frame): Parent frame.
                     provider (Provider): Provider.
+                    overwrite (bool): True to overwrite existing frame with same name (default: False).
```
apaletta3:
```suggestion
                    overwrite (bool): True to overwrite existing frame with same name. Defaults to False.
```

apaletta3 (approved):
Nice!

vishwa2710 (approved):
Looks great, I believe this should be a breaking change, as it changes the ABI. We are planning on doing a breaking release for OSTk soon so we can bundle it in with that. Let's update the description to be feat!

Issue #336, opened 2025-05-30
[fix] Issue title


jeremy-muon:
The Position::inFrame() function inserts cachedTransforms() into a map. It doesn't evict items from the cache normally, however, leading to significant leaking of memory. Here is a view of the memory still in use at process exit time of one test binary:

![Image](https://github.com/user-attachments/assets/2608d0ca-faa6-497f-8bd6-12ae408ae855)

**Steps to reproduce**

There are certain functions that cause an especially high number of insertions into this cache without evictions:

1. Call Model::CalculateRevolutionNumberAt() repeatedly.
2. Observe memory usage at program exit.

**Expected behavior**
The cache should not grow without bound.

**Potential suggestion**
There should be an upper limit on the size of the cache. Perhaps LRU caching plus a max size of ~10 MiB?

jeremy-muon:
Likely easiest quick and dirty fix:

1. Change [accessCachedTransform](https://github.com/open-space-collective/open-space-toolkit-physics/blob/3891b92f4fcbf7bf52c14a590998b6f588aea5e4/include/OpenSpaceToolkit/Physics/Coordinate/Frame/Manager.hpp#L46) to return a copy instead of a pointer.
2. Whenever [transformCache_](https://github.com/open-space-collective/open-space-toolkit-physics/blob/3891b92f4fcbf7bf52c14a590998b6f588aea5e4/include/OpenSpaceToolkit/Physics/Coordinate/Frame/Manager.hpp#L66) has too many entries, just clear it before adding a new one.

This should make it thread-safe, and performance hit is just an increase in cache 'misses' whenever things get big.

Naturally, an LRU cache or something would have higher performance, but it may not be worth the additional code complexity. I'll bet the most value from the cache is for immediately repeated accesses to the same transform, as opposed to access to the same frame many operations later.

lucas-bremond:
Thanks for reporting @jeremy-muon! Yes, definitely something we should fix ASAP.

Your suggestions make sense. Let us take a closer look and get back to you on this.

vishwa2710:
Hi @jeremy-muon . We've released 12.3.0 with the fix to clear the cache at a maximum size, we'll probably improve this in the future to do LRU/FIFO based caching. Thanks for pointing out this issue!

PR #351, opened 2025-12-03
fix: check transform cache size properly

alex-liang3:
Re-addresses [Issue 336](https://github.com/open-space-collective/open-space-toolkit-physics/issues/336).

The transform cache looks something like this:
```
GCRF:
    ITRF:
        instant1: transform1
        instant2: transform2
    TEME:
        instant1: transform1
        instant2: transform2
ITRF:
    GCRF:
        instant1: transform1
        instant2: transform2
```

Previously, we were only checking the size of the outermost layer, which usually has very few elements (noting the default cache size of 1000). Rather, we should be checking the amount of `<Instant, Transform>` pairs for a given pair of reference frames. 

With the default size of 1000, this reduces the memory footprint by about ~3.5x, and fixes the memory accumulation. 

This MR is just to fix the original intent; a follow-up MR will improve the speed of the FrameManager and likely refactor this anyway. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache management for coordinate frame transformations to prevent unnecessary data clearing and ensure more predictable cache behavior during operations.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/OpenSpaceToolkit/Physics/Coordinate/Frame/Manager.cpp
+++ b/src/OpenSpaceToolkit/Physics/Coordinate/Frame/Manager.cpp
@@ -126,15 +126,17 @@ void Manager::addCachedTransform(
 {
     const std::lock_guard<std::mutex> lock {mutex_};
 
-    if (transformCache_.size() >= maxTransformCacheSize_)
-    {
-        // TODO: Implement LRU or FIFO eviction strategy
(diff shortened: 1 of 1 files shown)


apaletta3 (approved):
Great stuff!

PR #378, opened 2026-07-01
perf: compute Julian dates arithmetically in Instant

vishwa2710:
First in a set of MRs to improve the propagation performance. Upon profiling, it was shown that the slowest part is the frame transformation (no surprises there). This is about ~50% of that slowdown currently. More MRs to follow.

## Summary

`Instant::getJulianDate` / `Instant::getModifiedJulianDate` previously round-tripped through `getDateTime`, which calls `std::mktime` + `gmtime_r` (touching timezone state under a lock) and then converts the calendar date **back** into a Julian date through a ~2.4e6-magnitude double, quantizing the result to ~20-40 us of time. They are now computed directly from the internal nanosecond count with a whole-day/remainder split.

The CIRF/TIRF/ITRF frame providers now call these accessors directly instead of `getDateTime(scale).getModifiedJulianDate()`. The IERS Finals2000A/BulletinA lookups (two `getModifiedJulianDate(UTC)` calls per query), SPICE time conversions, and NRLMSISE-00 all route through the same accessors and speed up for free.

## Why

Profiling a LEO numerical propagation (10x10 gravity + drag + Sun/Moon third-body) showed the per-derivative-evaluation cost is dominated by fixed frame-transform overhead rather than the spherical-harmonics math (0.76 us at 10x10). Of the 147 us GCRF->ITRF transform, ~75 us was mktime/gmtime_r datetime conversions and Julian-date round-trips - the TIRF leg cost 44 us while its actual math (iauEra00) costs 0.024 us.

## Measured impact (2000-call average, per fresh instant)

| Benchmark | Before | After |
|---|---|---|
| GCRF->ITRF getTransformTo | 147 us | ~62 us |
| Exponential density (LLA conversion path) | 142 us | ~85 us |
| **1-day LEO propagation, RK4 5 s, EGM96 10x10** | **6.8 s** | **~3.9 s** |

(A companion PR interpolating the IAU-2006/2000A X,Y,s series brings the transform down further, to ~13.5 us.)

## Accuracy

The new conversion is *more* accurate than the old one: the old JD round-trip quantized time to ~20-40 us, which aliased into ~300 uas of Earth-rotation-angle noise per transform (verified by dumping GCRF->ITRF quaternions over a year before/after: max difference 1.47e-9 rad, exactly the removed quantization noise). A consistency test sweeping +/-20 years across all time scales is included.

## Testing

- All existing Instant tests pass (exact-value expectations unchanged).
- New GetModifiedJulianDate_ConsistencyWithDateTime test (+/-20 years, all scales, pre/post-epoch).
- CIRF/TIRF/ITRF provider and Coordinate/Frame suites pass.
- **Downstream: all 49 astrodynamics Propagator/dynamics tests pass against this library.**

## Summary by CodeRabbit

* **Improvements**
  * Improved date and time conversions across supported time scales.
  * Increased precision and consistency when calculating Julian and Modified Julian Dates.
  * Improved nanosecond-level round-trip accuracy for date/time conversions.
  * Coordinate frame transformations now use more direct time calculations, improving reliability.

* **Bug Fixes**
  * Corrected minor differences in transformed ITRF position results.

* **Tests**
  * Added coverage for fractional-day calculations and date/time round trips across supported time scales.

### DIFF
--- a/src/OpenSpaceToolkit/Physics/Time/Instant.cpp
+++ b/src/OpenSpaceToolkit/Physics/Time/Instant.cpp
@@ -1,6 +1,7 @@
 /// Apache License 2.0
 
 #include <chrono>
+#include <cstdint>
 #include <iomanip>
 #include <stdlib.h>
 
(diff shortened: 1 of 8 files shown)


vishwa2710:
Added a second commit extending the same fix to `DateTime::getModifiedJulianDate()` — it had the identical precision flaw (computed as `getJulianDate() − 2400000.5`, quantizing to ~40 µs of time). Now computed directly from the calendar fields (JDN − 2400001 + from-midnight fraction), resolving ~0.6 µs and agreeing with the arithmetic `Instant` conversion.

Also in that commit:
- The `Position::InFrame` test's hardcoded GCRF→ITRF reference embedded ~0.65 mm of the removed time-quantization noise (its tolerance was 1e-8 m — 10 nm!). Regenerated the reference and set a millimeter tolerance, which reflects the value's real sensitivity to the Earth-orientation model/data.
- `ostk-format-cpp` line-wrapping on the first commit's files.

Full test suite: **721/722 pass, 1 environment-gated skip, 0 failures** (the 15 BulletinA failures seen locally earlier were missing LFS test fixtures, since materialized).

test/OpenSpaceToolkit/Physics/Coordinate/Position.test.cpp
```diff
@@ -210,8 +210,12 @@ TEST_F(OpenSpaceToolkit_Physics_Coordinate_Position, InFrame)
 
         const Position position_ITRF = position_GCRF.inFrame(Frame::ITRF(), Instant::J2000());
 
+        // Reference value regenerated after the arithmetic Julian-date conversion removed
+        // ~20-40 us of time quantization (~0.65 mm here) from the GCRF -> ITRF transform.
+        // The millimeter tolerance reflects the sensitivity of this value to the underlying
+        // Earth-orientation model and data rather than to floating-point noise.
```
vishwa2710:
```suggestion
```

src/OpenSpaceToolkit/Physics/Time/Instant.cpp
```diff
@@ -11,6 +12,46 @@
...
+
+// Low-level proleptic-Gregorian calendar algorithms, after Howard Hinnant's
+// "chrono-Compatible Low-Level Date Algorithms" (https://howardhinnant.github.io/date_algorithms.html)
+
+/// @brief Number of days from 1970-01-01 to the given civil date
+std::int64_t daysFromCivil(std::int32_t aYear, std::uint32_t aMonth, std::uint32_t aDay)
+{
+    aYear -= aMonth <= 2;
+    const std::int32_t era = (aYear >= 0 ? aYear : aYear - 399) / 400;
+    const std::uint32_t yearOfEra = static_cast<std::uint32_t>(aYear - era * 400);                 // [0, 399]
```
apaletta3, line 26: aren't you putting this in the global namespace? that's not what we want right? just want it to be scoped to this module right? Could also consider making it a private static method so that its not exposed at the API?

apaletta3 (commented):
Nice - LGTM pending hiding the helper functions from the API

src/OpenSpaceToolkit/Physics/Time/Instant.cpp
vishwa2710: These are in an anonymous namespace, and are therefore just scoped to this specific .cpp file.
IMO this is preferred to having private static methods as:
- private static methods are required to be defined in the header, forcing potentially unnecessary re-compilation when making changes
- extra code in the header class
- more readable, communicates to the reader that those functions are scoped to be only within that .cpp file (kinda like private methods in python within a file)
- cannot be used by friend classes

apaletta3: sounds good, I'm ok with that

apaletta3: I just didn't want them to exposed at the API, which they are not, so all good

apaletta3 (approved):

PR #379, opened 2026-07-01
perf: interpolate IAU 2006/2000A X, Y, s in CIRF provider

vishwa2710:
## Summary

`iauXys06a` — the full IAU 2006/2000A precession-nutation X, Y, s series (~1600 trig terms) — costs **~50 µs per evaluation** and dominates the GCRF↔CIRF (and therefore GCRF↔ITRF) transform. Its output varies smoothly in time: the shortest significant nutation periods are ~5 days. This PR evaluates the series on a uniform **0.25-day grid** and interpolates with a **centered 8-point Lagrange polynomial**, computing grid nodes on demand into a bounded, mutex-protected cache (file-local, no header/API change).

Direct per-call evaluation remains available via `OSTK_PHYSICS_COORDINATE_FRAME_PROVIDER_CIRF_XYS_INTERPOLATION=Disabled`.

## Accuracy

Dumped GCRF→ITRF quaternions at 400 instants over a year, interpolated vs. direct: max difference **1e-4 µas** (= 0.04 mm at LEO radius). For reference, the constant CIP offsets applied on top of the series are ~200 µas and IERS dX/dY corrections are O(100 µas) — the interpolation error is 2 orders of magnitude below the model's own uncertainty floor. Why so small: X, Y, s are dominated by polynomial (precession) terms plus periodic terms of ≥ ~5-day period; 8-point Lagrange on a 0.25-day grid reproduces a 5-day sinusoid to ~1e-5 relative.

## Measured impact (per fresh instant)

| Benchmark | Before #378 | After #378 | + this PR |
|---|---|---|---|
| GCRF→CIRF | 72 µs | 55 µs | **7.7 µs** |
| GCRF→ITRF | 147 µs | 62 µs | **13.5 µs** |
| 1-day LEO propagation (RK4 5 s, EGM96 10×10) | 6.8 s | 3.9 s | **1.9 s** |
| Same + drag + Sun/Moon third-body | 15.6 s | — | **9.4 s** |

Amortization: a 1-day propagation triggers ~8 `iauXys06a` node evaluations instead of ~34,560 direct calls.

For reference, this is also done in [Orekit](https://www.orekit.org/site-orekit-13.1.5/architecture/frames.html)

> Since Orekit uses the new paradigm for IERS frames, the IAU-2006 precession and IAU-2000A nutation model implemented are the complete model with thousands of luni-solar and planetary terms (1600 terms for the x components, 1275 components for the y component and 66 components for the s correction). Recomputing all these terms each time the CIRF frame is used would be really slow. Orekit therefore implements a caching/interpolation feature to improve efficiency. The shortest period for all the terms is about 5.5 days (it is related to one fifth of the moon revolution period). The pole motion is therefore quite smooth at the day or week scale. This implies that this motion can be computed accurately using a few reference points per day or week and interpolated between these points. The trade-off selected for Orekit implementation is to use eight points separated by four hours each. The resulting maximal interpolation error on the frame is about 4e-15 radians. The reference points are cached so the computation cost is roughly to perform two complete evaluations of the luni-solar and planetary terms per simulation day, and one interpolation per simulation step, regardless of the step size. This represents huge savings for steps shorter than one half day, which is the rule in most application (step sizes are mostly of the range of a few tens of seconds). Note that starting with Orekit 6.0, this caching feature is thread-safe.

## Summary by CodeRabbit

* **New Features**
  * Added Python access to the CIRF coordinate-frame provider.
  * Added optional X/Y/s interpolation for faster CIRF transform calculations, disabled by default.
  * Added runtime controls to enable or disable interpolation and clear its cache.
  * Added interpolation caching with improved accuracy and thread-safe behavior.

* **Documentation**
  * Documented the interpolation environment-variable setting and its default value.

* **Bug Fixes**
  * Removed stray diagnostic output from an IERS test.


vishwa2710:
Additional data: 1-day vs 7-day propagation (RK4 5 s, EGM96 10×10, Release builds throughout), interpolation vs direct `iauXys06a`:

| Propagation | interp ON | interp OFF | speedup |
|---|---|---|---|
| 1 day | 0.97 s | 2.89 s | 3.0× |
| 7 days | 4.7 s | 17.9 s | 3.8× |

Steady-state marginal cost: ~634 ms/day vs ~2517 ms/day — the advantage grows with duration since a 7-day propagation computes only **31 grid nodes** (~1.6 ms of series evaluation) versus 241,920 direct calls (~12.7 s).

Final-state agreement (interp vs direct, same scenario): **4 µm after 1 day, 0.23 mm after 7 days** — per-transform error is bounded (~1 µas) and does not accumulate secularly.

include/OpenSpaceToolkit/Physics/Coordinate/Frame/Provider/CIRF.hpp
```diff
@@ -69,6 +73,47 @@ class CIRF : public Provider
...
+    /// @ref getTransformAt), evaluated either directly via the SOFA series or by
+    /// interpolating a cached grid of the series (see @ref IsXysInterpolationEnabled).
+    ///
+    /// @param [in] aModifiedJulianDate_TT A Modified Julian Date, in the TT scale
+    /// @param [out] x The CIP X coordinate [rad]
+    /// @param [out] y The CIP Y coordinate [rad]
+    /// @param [out] s The CIO locator s [rad]
+    /// @param [in] interpolate If true, interpolate a cached grid of the series; otherwise evaluate it directly
+    static void ComputeCIPCoordinates(
+        const Real& aModifiedJulianDate_TT, double& x, double& y, double& s, const bool interpolate
```
apaletta3: why choose this signature instead of our more familiar const input (can still be by reference) returning const output signature? for performance? or because you would have to encapsulate the x, y, s in an intermediate class/struct since cpp doesn't support multiple return types?

If you'd like to keep it in this format, can we perhaps hide the method from the API? Does it need to be exposed?

bindings/python/src/OpenSpaceToolkitPhysicsPy/Coordinate/Frame/Provider/CIRF.cpp
```diff
@@ -0,0 +1,131 @@
...
+                    instant (Instant): An instant.
+
+                Returns:
+                    Transform: The transform from GCRF to CIRF at the given instant.
+            )doc"
+        )
+
+        .def_static(
+            "compute_cip_coordinates",
+            [](const Real& aModifiedJulianDate_TT, const bool interpolate) -> std::tuple<double, double, double>
```
apaletta3: you have to wrap the function in python because of the non-const references under the hood? If so can you add a comment explaining that?

include/OpenSpaceToolkit/Physics/Coordinate/Frame/Provider/CIRF.hpp
```diff
@@ -69,6 +73,47 @@ class CIRF : public Provider
     /// @param [in] anInstant An instant
     /// @return The transform from GCRF to CIRF at the given instant
     virtual Transform getTransformAt(const Instant& anInstant) const override;
+
+    /// @brief Compute the IAU 2006/2000A CIP X, Y and CIO locator s at a given instant.
```
apaletta3: nit: can you define some of these acronyms somewhere? (wherever it makes the most sense)

include/OpenSpaceToolkit/Physics/Coordinate/Frame/Provider/CIRF.hpp
```diff
@@ -69,6 +73,47 @@ class CIRF : public Provider
...
+    /// Defaults to disabled, unless the environment variable
+    /// OSTK_PHYSICS_COORDINATE_FRAME_PROVIDER_CIRF_XYS_INTERPOLATION is set to "Enabled" or "True".
+    /// Can be overridden at runtime via @ref SetXysInterpolationEnabled.
+    ///
+    /// @return True if X, Y, s interpolation is enabled
+    static bool IsXysInterpolationEnabled();
+
+    /// @brief Enable or disable X, Y, s interpolation at runtime.
+    ///
+    /// Overrides the environment-variable default. Primarily intended for testing and benchmarking.
```
apaletta3: it overrides it statically - aka for the whole lifetime of the program, correct? If so can you clarify this in the cpp/python docs?

apaletta3 (commented):
Nice!

Left a couple small questions in the MR

Overall comment/question, though, is: The current static/mutexed implementation of this cached interpolation is a little complex and more involved. Was this the only way to achieve the performance increase or was there another more "instance" based way to do it that was considered?

include/OpenSpaceToolkit/Physics/Coordinate/Frame/Provider/CIRF.hpp
vishwa2710: So the pass by reference is for performance, (don't have to create a new struct etc.)

As for making it private, we could, but it was the only way to test the interpolation vs non-interpolation cleanly. Let me see if I can think of something. I agree it's not a very useful method to expose.

vishwa2710: Okay I think this is a bit cleaner. Let me know what you think.

apaletta3 (approved):
Nice! LGTM

PR #381, opened 2026-07-02
perf: convert Instant/DateTime calendar conversions to pure arithmetic

vishwa2710:
Second MR based on date time conversion performance (and accuracy as bonus)

## Summary

`Instant::getDateTime` and the `Instant::DateTime` factory previously went through `std::mktime` + `gmtime_r` — which consult libc timezone state under a lock — plus chrono round-trips. Both are now pure integer arithmetic using [Howard Hinnant's civil-date algorithms](https://howardhinnant.github.io/date_algorithms.html) (128-bit intermediates, since the nanosecond count spans more than the signed 64-bit range).

## Measured (20,000-call average, same machine)

| Conversion | Before | After |
|---|---|---|
| `Instant::DateTime` (factory) | 13.53 µs | **0.20 µs** (~69×) |
| `Instant::getDateTime` | 6.58 µs | **0.14 µs** (~46×) |
| Full physics test-suite wall time | 79 s | **39 s** |

These sit on hot paths: NRLMSISE-00 makes several calendar conversions per density evaluation, and string formatting, parsing, and space-weather lookups all route through them.

## Behavior notes

- The `[1970, 2554]` year guard on the factory is preserved exactly (it exists for the `Uint64` nanosecond range; the lower bound was also a `mktime` portability limit and could now be relaxed toward the `Date` range in a follow-up if desired).
- The old `getDateTime` derived sub-second fields by truncating chrono durations toward zero — incorrect for hypothetical pre-1970 instants; the floor-division arithmetic is uniformly correct.
- No timezone dependence remains (the old path was correct only because the two `mktime` calls' offsets cancelled with `tm_isdst=0`).

## Testing

- New `DateTime_RoundTrip` test: nanosecond-resolution calendar round-trips at the range edges (1970-01-01, 2554-12-31), leap day, year boundaries, across all time scales.

### DIFF
--- a/src/OpenSpaceToolkit/Physics/Time/Instant.cpp
+++ b/src/OpenSpaceToolkit/Physics/Time/Instant.cpp
@@ -1,6 +1,7 @@
 /// Apache License 2.0
 
 #include <chrono>
+#include <cstdint>
 #include <iomanip>
 #include <stdlib.h>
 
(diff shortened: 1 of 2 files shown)


test/OpenSpaceToolkit/Physics/Coordinate/Position.test.cpp
```diff
@@ -210,6 +210,10 @@ TEST_F(OpenSpaceToolkit_Physics_Coordinate_Position, InFrame)
 
         const Position position_ITRF = position_GCRF.inFrame(Frame::ITRF(), Instant::J2000());
 
+        // Reference value regenerated after the arithmetic Julian-date conversion removed
+        // ~20-40 us of time quantization (~0.65 mm here) from the GCRF -> ITRF transform.
+        // The millimeter tolerance reflects the sensitivity of this value to the underlying
+        // Earth-orientation model and data rather than to floating-point noise.
```
vishwa2710:
```suggestion
```

PR #392, opened 2026-08-12
feat: add Earth atmospheric model constructor

vishwa2710:
Constructing an Earth atmospheric model with a non-default Earth frame previously required spelling out the input data type and the three constant flux/geomagnetic index values first, since the frame sits in the sixth position. Add an overload taking the frame as the second argument so the remaining parameters can keep their defaults.

This makes it easy to create an earth atmospheric model that uses a Spice based frame (with the other sensible defaults):
```
EarthAtmosphericModel(
    type=EarthAtmosphericModel.Type.NRLMSISE00,
    earth_frame=ephemeris.access_frame(),
)
```

## Summary by CodeRabbit

* **New Features**
  * Added an Earth atmospheric model constructor that accepts an Earth reference frame.
  * Optional atmospheric model parameters now include documented default values.
  * CSSI space-weather data is used by default when no input type is specified.

* **Bug Fixes**
  * Improved support for configuring atmospheric models with Earth-frame data.

* **Tests**
  * Added coverage for new constructor options, default settings, invalid inputs, and density calculations.


apaletta3 (approved):
Nice

PR #400, opened 2026-09-10
chore: deprecate date and time class mutable setters

### LINKED ISSUE #401: [feat] Add a replaceHour method
**Is your feature request related to a problem? Please describe.**
https://github.com/open-space-collective/open-space-toolkit-physics/pull/400#pullrequestreview-5169925149

apaletta3:
We will remove these in the future, because they aren't really used/useful and prevent us from making these types hashable.

Instead, users can use methods that copy the underlying value and replace it with their preferred value (just like in python datetime.replace) - this way the original object is not mutated.


vishwa2710 (commented):
I think it's worth adding `replaceHour` etc. methods so that it's convenient for the user to update 1 value (it returns a new object rather than sets in place).

apaletta3:
> I think it's worth adding `replaceHour` etc. methods so that it's convenient for the user to update 1 value (it returns a new object rather than sets in place).

I can add a ticket for that and put it in a separate MR

src/OpenSpaceToolkit/Physics/Time/Date.cpp
```diff
@@ -385,34 +401,39 @@ Date::Date()
 {
 }
 
-void Date::ValidateDate(Uint16 aYear, Uint8 aMonth, Uint8 aDay)
+void Date::ValidateDate(const Integer& aYear, const Integer& aMonth, const Integer& aDay)
```
apaletta3, line 404: these had to be changed to take in `Integer` instead of `Uint8/Uint16`. Without this, a value like replace(day=257) would have silently wrapped to 1 when narrowed to Uint8 and been accepted, and negatives would have reported nonsense values in the error message. 

Now out-of-range and negative inputs throw with the actual value in the message, and the range messages are unchanged for valid-typed inputs.

include/OpenSpaceToolkit/Physics/Time/Time.hpp
```diff
@@ -321,12 +361,12 @@ class Time
...
-    static void ValidateSecond(Uint8 aSecond);
-    static void ValidateMillisecond(Uint16 aMillisecond);
-    static void ValidateMicrosecond(Uint16 aMicrosecond);
-    static void ValidateNanosecond(Uint16 aNanosecond);
+    static void ValidateHour(const Integer& anHour);
+    static void ValidateMinute(const Integer& aMinute);
+    static void ValidateSecond(const Integer& aSecond);
+    static void ValidateMillisecond(const Integer& aMillisecond);
+    static void ValidateMicrosecond(const Integer& aMicrosecond);
+    static void ValidateNanosecond(const Integer& aNanosecond);
```
vishwa2710: I wouldn't convert these to `Integer`, they're stored as Uint8 and Uint16 respectively under the hood.

vishwa2710 (approved):

PR #403, opened 2026-09-19
feat: allow the frame transform cache size to be set at runtime

vishwa2710:
## What

The transform cache bound could only be chosen through `OSTK_PHYSICS_FRAME_MANAGER_MAX_TRANSFORM_CACHE_SIZE`, read once when the `Manager` singleton is constructed. A library that knows how many distinct instants it is about to evaluate had no way to size the cache for that work, and a caller who cannot set environment variables before process start had no way to reach it at all.

Adds `getMaxTransformCacheSize` / `setMaxTransformCacheSize`, both taking the existing mutex, plus Python bindings. The environment variable still supplies the initial value, so nothing changes for existing callers.

Lowering the bound drops any frame pair already holding more than the new value, rather than leaving it oversized until the next insertion into that particular pair notices.

## Why it matters

The cache is keyed by frame pair and, within a pair, by instant. On reaching the bound, the instants cached for that pair are dropped **wholesale** rather than evicted one at a time:

```cpp
if (transformCacheToFrameIt->second.size() >= maxTransformCacheSize_)
{
    // Clear instants for this frame pair only
    // TBI: Improve caching strategy, perhaps LRU.
    transformCacheToFrameIt->second.clear();
}
```

So work that walks a long span of distinct instants does not degrade gradually as the cache fills — it falls off a cliff once it passes the bound, because after each drop it repopulates entries it will never ask for again.

Measured in `open-space-toolkit-astrodynamics`, one ground station against one satellite, sweeping the analysis window at a fixed 1-minute step:

| Window | Default (1000) | 100 000 | Speedup |
| --- | ---: | ---: | ---: |
| 1 day | 11.8 ms | 12.1 ms | 1.0× |
| 4 days | 48.0 ms | 46.8 ms | 1.0× |
| 5 days | 222.4 ms | 57.8 ms | 3.8× |
| 7 days | 406.2 ms | 82.0 ms | 5.0× |
| 14 days | 834.3 ms | 163.9 ms | **5.1×** |

Scaling is exactly linear at the larger bound; all of the super-linearity below it is the cache turning over. Profiling attributes it where you would expect: `Frame::getTransformTo` is 73–99% of instructions in that workload, and `iauNut00a` — the 1365-term IAU 2000A nutation series reached through `CIRF::getTransformAt` — is 29–59% on its own.

## Why a setter and not a new default

A larger cache is not free to maintain. On workloads that reuse a small set of instants across many queries — many fixed ground stations sharing one time grid, for instance — raising it measured *marginally slower*:

| Workload | Default | 100 000 |
| --- | ---: | ---: |
| 100 fixed targets, 1 week, elevation criterion | 232 ms | 242 ms |
| 100 fixed targets, 1 day, elevation criterion | 33.9 ms | 34.8 ms |
| GCRF-output trajectory model, 1 target, 2 weeks | 2378 ms | 339 ms (7.0×) |

What decides it is how many *distinct* instants a run evaluates, not how long it is. The right size depends on the work, so this exposes the knob rather than guessing a better constant.

The existing `TBI: Improve caching strategy, perhaps LRU` note still stands, and is the better long-term fix: an LRU policy would remove the cliff rather than let callers step around it. Happy to take that on separately if you'd like it.

## Testing

Adds a C++ test covering the accessor, rejection of zero, the drop at the bound, that raising preserves what is cached, and that lowering evicts immediately; plus a Python test for the bindings.

> **I was not able to run the physics test suite for this change.** Building this repo needs SOFA and CSPICE, which are `FIND_PACKAGE` dependencies I could not provision in my environment. What I did verify: all three changed translation units — `Manager.cpp`, `Manager.test.cpp` and the pybind11 binding — compile cleanly (`-fsyntax-only`) against the current headers. The measurements quoted above were taken by driving the existing environment variable through the astrodynamics benchmark suite, which exercises the same bound the setter writes to. Please treat CI as the real gate here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01T8xLC4jVuShuAQPp81rJ4Y

---
_Generated by [Claude Code](https://claude.ai/code/session_01T8xLC4jVuShuAQPp81rJ4Y)_

## Summary by CodeRabbit

* **New Features**
  * Added controls for viewing and setting the maximum number of cached transforms retained for each frame pair.
  * Cache limits must be greater than zero.
  * Lowering the limit immediately removes cached frame-pair entries that exceed the new limit.

* **Bug Fixes**
  * Improved cache management when transform limits are changed at runtime.


bindings/python/src/OpenSpaceToolkitPhysicsPy/Coordinate/Frame/Manager.cpp
```diff
@@ -136,6 +138,32 @@ inline void OpenSpaceToolkitPhysicsPy_Coordinate_Frame_Manager(pybind11::module&
...
+        .def(
+            "set_max_transform_cache_size",
+            &Manager::setMaxTransformCacheSize,
+            arg("max_transform_cache_size"),
+            R"doc(
+                Set the maximum number of cached transforms held per frame pair.
+
+                Work that walks a long span of distinct instants will pass the default bound, past which very
+                nearly every lookup misses. Sizing the cache to the number of distinct instants a run evaluates
+                avoids that. Lowering the bound drops any pair already holding more than it.
```
vishwa2710:
```suggestion
```

include/OpenSpaceToolkit/Physics/Coordinate/Frame/Manager.hpp
```diff
@@ -135,6 +135,37 @@ class Manager
...
+    /// @brief Set the maximum number of cached transforms held per frame pair.
+    ///
+    /// @details The cache is keyed by frame pair and, within a pair, by instant. This bounds the number of
+    /// instants held for each pair; once a pair reaches the bound, the instants cached for it are dropped and
+    /// repopulated from scratch. Work that walks a long span of distinct instants — access generation, a
+    /// propagation reported at fine cadence — will pass that bound, and past it very nearly every lookup misses.
+    /// Sizing the cache to the number of distinct instants a run evaluates avoids that.
+    ///
+    /// A larger cache is not free: it is worth raising when a run evaluates many distinct instants, and worth
+    /// nothing when it reuses a small set of them. Prefer measuring over raising it on principle.
```
vishwa2710:
```suggestion
    /// A larger cache is not free: it is trading memory for performance.
```

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Manager.cpp
```diff
@@ -178,6 +178,38 @@ void Manager::addCachedTransform(
...
+    {
+        throw ostk::core::error::runtime::Wrong("Maximum transform cache size");
+    }
+
+    const std::lock_guard<std::mutex> lock {mutex_};
+
+    maxTransformCacheSize_ = aMaxTransformCacheSize;
+
+    // Lowering the bound leaves pairs that already hold more than it. Drop those now, rather than waiting for
+    // the next insertion into each one to notice.
```
apaletta3: why did you choose to do it this way? Perhaps you can add a rationale for that here?

apaletta3 (approved):

src/OpenSpaceToolkit/Physics/Coordinate/Frame/Manager.cpp
vishwa2710: Cause currently our cache is not an **L**east **R**ecently **U**pdated (LRU) cache, it just clears it after we cross the limit.

